PR #13, opened 2019-07-23
Add unit tests to check convergence on the Max Entropy problem

rth:
Add a few unit tests that check convergence, adapted from [scipy.optimize.tests.test_optimize](https://github.com/scipy/scipy/blob/8e40c067b64016dc344c537655f56454fd1788ba/scipy/optimize/tests/test_optimize.py#L51).

The following tests were added,
```
test tests::test_max_entropy_bfgs ... ok
test tests::test_max_entropy_lbfgs_hagerzhang ... ok
test tests::test_max_entropy_dfp ... ok
test tests::test_max_entropy_newton_cg ... ok
test tests::test_max_entropy_lbfgs_morethuente ... ok
test tests::test_max_entropy_steepest_descent ... ok
```
so not all optimizer are for now included.

I have put these tests in `src/tests.rs` not sure if it's the best location. Maybe `src/solvers/tests.rs`? The idea was keep these in one place since they are "common tests" and then solver specific tests could go into separate files.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1,0 +1,145 @@
+// Copyright 2019
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+#![allow(non_snake_case)]
+
+use ndarray::prelude::*;
+use ndarray::{Array1, Array2};
+
+use crate::prelude::*;
+use crate::solver::gradientdescent::SteepestDescent;
+use crate::solver::linesearch::{HagerZhangLineSearch, MoreThuenteLineSearch};
+use crate::solver::newton::NewtonCG;
+use crate::solver::quasinewton::{BFGS, DFP, LBFGS};
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Default, Serialize, Deserialize, Debug)]
+struct MaxEntropy {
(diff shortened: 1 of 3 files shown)


src/tests.rs
```diff
@@ -0,0 +1,145 @@
...
+     test_max_entropy_lbfgs_morethuente: LBFGS::new(MoreThuenteLineSearch::new(), 10),
+     test_max_entropy_lbfgs_hagerzhang: LBFGS::new(HagerZhangLineSearch::new(), 10),
+     test_max_entropy_bfgs: BFGS::new(Array2::eye(3), MoreThuenteLineSearch::new()),
+     test_max_entropy_dfp: DFP::new(Array2::eye(3), MoreThuenteLineSearch::new()),
+     test_max_entropy_newton_cg: NewtonCG::new(MoreThuenteLineSearch::new()),
+     test_max_entropy_steepest_descent: SteepestDescent::new(MoreThuenteLineSearch::new()),
+}
+
+#[test]
+fn test_lbfgs_func_count() {
```
rth: This tests is somewhat redundant with the parallelization above, but the point here was to check the number of function and gradient calls.

stefan-k:
This crate desperately needs more tests. Thanks a lot!

> I have put these tests in `src/tests.rs` not sure if it's the best location. Maybe `src/solvers/tests.rs`? 

Both locations sound good to me. 

> The idea was keep these in one place since they are "common tests" and then solver specific tests could go into separate files.

I agree.

stefan-k:
This looks pretty good to me, is it ready to be merged?

rth:
Thanks for the review. Yes, good to go on my side, I would rather open another PR for other changes.

Issue #1, opened 2019-02-04
Road to 1.0.0


stefan-k:
This is a long-term issue in order to keep these things in mind.

- **Naming** *(crate aligns with Rust naming conventions)*
  - [ ] Casing conforms to RFC 430 ([C-CASE])
  - [ ] Ad-hoc conversions follow `as_`, `to_`, `into_` conventions ([C-CONV])
  - [ ] Getter names follow Rust convention ([C-GETTER])
  - [ ] Methods on collections that produce iterators follow `iter`, `iter_mut`, `into_iter` ([C-ITER])
  - [ ] Iterator type names match the methods that produce them ([C-ITER-TY])
  - [ ] Feature names are free of placeholder words ([C-FEATURE])
  - [ ] Names use a consistent word order ([C-WORD-ORDER])
- **Interoperability** *(crate interacts nicely with other library functionality)*
  - [ ] Types eagerly implement common traits ([C-COMMON-TRAITS])
    - `Copy`, `Clone`, `Eq`, `PartialEq`, `Ord`, `PartialOrd`, `Hash`, `Debug`,
      `Display`, `Default`
  - [ ] Conversions use the standard traits `From`, `AsRef`, `AsMut` ([C-CONV-TRAITS])
  - [ ] Collections implement `FromIterator` and `Extend` ([C-COLLECT])
  - [ ] Data structures implement Serde's `Serialize`, `Deserialize` ([C-SERDE])
  - [ ] Types are `Send` and `Sync` where possible ([C-SEND-SYNC])
  - [ ] Error types are meaningful and well-behaved ([C-GOOD-ERR])
  - [ ] Binary number types provide `Hex`, `Octal`, `Binary` formatting ([C-NUM-FMT])
  - [ ] Generic reader/writer functions take `R: Read` and `W: Write` by value ([C-RW-VALUE])
- **Macros** *(crate presents well-behaved macros)*
  - [ ] Input syntax is evocative of the output ([C-EVOCATIVE])
  - [ ] Macros compose well with attributes ([C-MACRO-ATTR])
  - [ ] Item macros work anywhere that items are allowed ([C-ANYWHERE])
  - [ ] Item macros support visibility specifiers ([C-MACRO-VIS])
  - [ ] Type fragments are flexible ([C-MACRO-TY])
- **Documentation** *(crate is abundantly documented)*
  - [ ] Crate level docs are thorough and include examples ([C-CRATE-DOC])
[...]

PR #14, opened 2019-07-25
Derive Clone trait for all optimizers

rth:
Makes sure all optimizer derive the `Clone` trait.

The only exception is `ParticleSwarm` because it sounds like `Callback<T>` does not implement `Clone` and I have not investigated further.

Something I have run into in `pyargmin` when changing solvers.

Helps toward "Interoperability" in https://github.com/argmin-rs/argmin/issues/1

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -38,6 +38,5 @@ macro_rules! test_trait_impl {
                 assert_clone::<$t>();
             }
         }
-
     };
 }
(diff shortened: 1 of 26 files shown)


stefan-k:
This is a good idea!

I think we can ignore `ParticleSwarm` for now because as far as I know the callback is only for plotting and I think @jjbayer is implementing an `Observer` for the plotting instead (in the `visualizer` branch). If I'm not mistaken this would make the callback obsolete.

PR #22, opened 2019-12-18
Brent's method

xemwebe:
As discussed in the Issues, sections, I have implemented Brent's method and this is the pull request.


stefan-k:
Thanks for this contribution! I've decided to respond to your comments here instead of in issue #20, I hope this is alright with you.

> However, a few questions remain: I haven't seen any place where I could set the required precision. I would assume that these would be ideally handled by the Executor, automatically stopping the process once the required precision is reached, though this would require some input from the solver.

You are right, the required precision could be part of `IterState` and exposed to the user via the `Executor`, just like `max_iters`. You can define custom stopping criteria via the `terminate` method of the `Solver` trait ([see here](https://github.com/argmin-rs/argmin-core/blob/7dc429c0da9592ae9de8ae632454ea3ccc76517a/src/lib.rs#L172)). This will be executed at the [beginning of each iteration](https://github.com/argmin-rs/argmin-core/blob/7dc429c0da9592ae9de8ae632454ea3ccc76517a/src/executor.rs#L148). Alternatively it could also be implemented [here](https://github.com/argmin-rs/argmin-core/blob/7dc429c0da9592ae9de8ae632454ea3ccc76517a/src/lib.rs#L157) which would then apply for all solvers, but I'm not sure if this is appropriate. If I remember correctly, some solvers require to do such checks somewhere in the middle of an iteration, which would then not be possible.

> Also, Brent's method require to store some state in between iterations, which is now stored in the Brent struct, since IterState does not seem to be appropriate for this purpose.

This is absolutely fine, `IterState` is for storing state which applies to all (or at least most) solvers. I'm open for suggestions if you think other fields should be included. Maybe an additional `HashMap` could be useful.

src/solver/brent/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+/// Brent's method
+///
+/// A root-finding algorithm combining the bisection method, the secant method
+/// and inverse quadratic interpolation. It has the reliability of bisection
+/// but it can be as quick as some of the less-reliable methods.
+///
+/// # References:
+/// https://en.wikipedia.org/wiki/Brent%27s_method
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Brent {
```
stefan-k: Could you please comment the individual fields?

src/solver/brent/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+                eff_tol.abs()
+            } else {
+                -eff_tol.abs()
+            };
+        }
+
+        self.fb = op.apply(&self.b)?;
+        Ok(ArgminIterData::new().param(self.b).cost(self.fb.abs()))
+    }
+}
```
stefan-k: Please add a test for the implemented traits, [for instance similar to this](https://github.com/argmin-rs/argmin/blob/5f33e8932c94b010ae62612eb956110d0d0436bf/src/solver/conjugategradient/cg.rs#L144).

src/solver/brent/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+        self.fc = self.fb;
+        Ok(Some(
+            ArgminIterData::new().param(self.b).cost(self.fb.abs()),
+        ))
+    }
+
+    fn next_iter(
+        &mut self,
+        op: &mut OpWrapper<O>,
+        // Brent maintains its own state
```
stefan-k: `IterState` will automatically track your current parameter vector and the current best parameter vector and their corresponding cost function values. Maybe you could reduce some of the state of the `Brent` struct with this (if you want, I personally think this is a matter of taste).

src/solver/brent/mod.rs
xemwebe: I have added the comments.

src/solver/brent/mod.rs
xemwebe: >You can define custom stopping criteria via the `terminate` method of the `Solver` trait ([see here](https://github.com/argmin-rs/argmin-core/blob/7dc429c0da9592ae9de8ae632454ea3ccc76517a/src/lib.rs#L172)). This will be executed at the [beginning of each iteration](https://github.com/argmin-rs/argmin-core/blob/7dc429c0da9592ae9de8ae632454ea3ccc76517a/src/executor.rs#L148). Alternatively it could also be implemented [here](https://github.com/argmin-rs/argmin-core/blob/7dc429c0da9592ae9de8ae632454ea3ccc76517a/src/lib.rs#L157) which would then apply for all solvers, but I'm not sure if this is appropriate. If I remember correctly, some solvers require to do such checks somewhere in the middle of an iteration, which would then not be possible.

I have already used the possibility to set a custom termination reason, so that should be fine. 
And I agree, controlling the accuracy by the `Executor` might not work in all cases.

src/solver/brent/mod.rs
xemwebe: I considered that, but the fields `prev_param`, `prev_best_param`, etc. doesn't really fit the meaning and it seemed to be an abuse to use them for other purpose. I also wasn't sure if these values might be set automatically by the the `Executor`, in which case it wouldn't work anyway.

stefan-k: I've made similar experiences with some solvers. The `IterState` design works for most solvers and adds a lot of convenience, but for some solvers it is just not appropriate.

src/solver/brent/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+/// https://en.wikipedia.org/wiki/Brent%27s_method
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Brent {
+    /// required relative accuracy
+    tol: f64,
+    /// left or right boundary of current interval
+    a: f64,
+    /// currently proposed best guess
+    b: f64,
+    // left or right boundary of current interval
```
stefan-k:
```suggestion
    /// left or right boundary of current interval
```

src/solver/brent/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+    c: f64,
+    // helper variable
```
stefan-k:
```suggestion
    /// helper variable
```

src/solver/brent/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+    d: f64,
+    // helper variable
```
stefan-k:
```suggestion
    /// helper variable 
```

src/solver/brent/mod.rs
stefan-k: These changes should fix the linter errors in the CI.

src/solver/brent/mod.rs
xemwebe: fixed

stefan-k:
Great, thank you once again for this contribution!!

Issue #29, opened 2020-03-27
Discussion about the structure of the project


stefan-k:
When I started this project, I wanted it to be as modular as possible, which is why I've decided to split the project into a "core" part which holds all the "infrastructure" and a part where all the solvers are implemented. The idea was that others, who wanted to implement solvers based on the principles introduced here, would not need to import the entire project. However, this has lead to an awkward development-experience: In order to develop locally, all repositories have to be pulled and `argmin`'s `Cargo.toml` needs to be adapted to point to the local crates. 

Does anyone have a suggestion how this could be solved better? Workspaces may be an option but I'm not sure if they sufficiently improve the situation. I'm tending towards just moving the core part into `argmin`...
I would still keep `testfunctions`, `finitediff`, `modcholesky` and `pyargmin` separated, as they seem sufficiently self-contained.

rth:
> However, this has lead to an awkward development-experience: In order to develop locally, all repositories have to be pulled and argmin's Cargo.toml needs to be adapted to point to the local crates.

Thanks @stefan-k I can confirm. I was hesitant to open an issue about it last week, but I can see the initial motivation as well. Another non ideal side effect of this approach, is that dependencies are split between 2 crates and the logic (e.g. of using ndarray with blas features) is a bit more complex as a result, particularly when used from say from pyargmin (i.e. pyargmin->argmin->argmin-core). Also searching for structs/traits in the error messages can be confusing as sometimes they are in argmin-core instead of argmin (and therefore not present in the documentation).

> I'm tending towards just moving the core part into argmin...
I would still keep testfunctions, finitediff, modcholesky and pyargmin separated, as they seem sufficiently self-contained.

+1 that would be useful I think. I don't have better suggestions in any case.

stefan-k:
Thanks for your input! I will try to find some time this weekend for this.

PR #30, opened 2020-03-29
Moved `argmin_core` crate into `core` module

stefan-k:
This moves all code from `argmin_core` into the `core` module. This will hopefully substantially improve the development experience. With this PR, the `argmin_core` crate becomes obsolete. 

See discussion in #29.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,28 +17,40 @@ exclude = [
 ]
 
 [dependencies]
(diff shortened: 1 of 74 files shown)


stefan-k:
I'm a bit confused because the travis builds are running, but for some reason don't show up here. I'll wait a bit before merging.

Issue #28, opened 2020-03-15
Adjust the stopping criterion


rth:
Currently the stopping criterion tolerance is often hard coded. For instance in LBFGS,
https://github.com/argmin-rs/argmin/blob/46f43cd753110cd853c74e190fe2dfcf145a5c6e/src/solver/quasinewton/lbfgs.rs#L166-L172

the stopping criterion for the gradient is `std::f64::EPSILON.sqrt() ~ 1e-8` while in practical machine learning application 1e-4 could be sufficient, as performance matter more than perfect convergence .

So the question is what API would allow for these stopping conditions to be set by the user. For instance, it could be some combination of,
 a) add `tol` to the struct initialization in `new`. That's probably not ideal, as for instance LBFGS would require 2 tolerances, and always providing them manually is not great.
 b) add `tol` to the struct but not as an argument in `new` method
 c) implement a `Default` trait with current values.
 d) some form of builder pattern or a method e.g. `with_tol`.

For now I would likely tend to propose b) with d) using a `with_tol` or just `tol` method (which also would be backward compatible), and I'm looking for some feedback. I'm not sure what would best generalize to different solvers. Also ideally the solution should also be consistent with [Executor.max_iters](https://argmin-rs.github.io/argmin/argmin/prelude/executor/struct.Executor.html#method.max_iters) approach.

stefan-k:
I thought about this problem in the past and couldn't come up with a totally satisfactory solution. Such tolerances may have slightly different meanings for different solvers, therefore a "general" solution at the level of the `Executor` (like `max_iters`) may cause confusion.

The downside of b+d is that it would be a bit inconsistent with the `max_iters`/`target_cost`/... approach. Nevertheless I personally would still also go for your proposal of b) and d) because it is probably more flexible.
At this point I have do admit I don't care so much about backwards compatibility yet.

In the meantime I started to like the prefix `with_` for such methods. Maybe we should change this wherever it makes sense? What do you think?

stefan-k:
Closing. All tolerances can now be set by the user.

PR #37, opened 2020-04-06
Exposed stopping criterion tolerances to user

stefan-k:
This PR makes the tolerances in the stopping criterion of L-BFGS accessible to the user (see #28). 

@rth, is this how you imagined it? I would appreciate it if you could give this a quick review.

### DIFF
--- a/src/solver/quasinewton/lbfgs.rs
+++ b/src/solver/quasinewton/lbfgs.rs
@@ -36,9 +36,9 @@ pub struct LBFGS<L, P> {
     s: VecDeque<P>,
     /// y_{k-1}
     y: VecDeque<P>,
-    /// Stopping criterion tolerance for precision
-    tol_precision: f64,
-    /// Stopping criterion tolerance for change in cost
+    /// Tolerance for the stopping criterion based on the change of the norm on the gradient
+    tol_grad: f64,
+    /// Tolerance for the stopping criterion based on the change of the cost stopping criterion
     tol_cost: f64,
 }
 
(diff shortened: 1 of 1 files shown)


src/solver/quasinewton/lbfgs.rs
```diff
@@ -46,8 +50,22 @@ impl<L, P> LBFGS<L, P> {
             m,
             s: VecDeque::with_capacity(m),
             y: VecDeque::with_capacity(m),
+            tol_precision: std::f64::EPSILON.sqrt(),
+            tol_cost: std::f64::EPSILON,
         }
     }
+
+    /// Sets tolerance for precision stopping criterion
+    pub fn with_tol_precision(mut self, tol_precision: f64) -> Self {
```
rth, line 59: I think it would help to be more explicit that this tolerance is applied on the gradient. E.g. in [scipy they have a pgtol](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin_l_bfgs_b.html) for projected gradient tolerance,
>  The iteration will stop when max{|proj g_i | i = 1, ..., n} <= pgtol where pg_i is the i-th component of the projected gradient.

How about `with_tol_grad`? I realize this makes it a bit more difficult to have common API for tolerance across solvers. If you want one main tolerance, maybe `with_tol` and make it very explicit in the docsting that it's taken on the gradient? I just find the word combination "tolerance precision" confusing.

stefan-k: Good point, thanks. I've opted for `with_tol_grad`. I've come to think that trying to have a common API across solvers may be counter-productive as tolerances often mean slightly different things for different solvers.

PR #38, opened 2020-04-06
Removed need for unwrap in cstep MoreThuente LineSearch

MattBurn:
Sometimes when solving a system with LBFGS and the MoreThuente linesearch, my optimisation would panic when unwrapping a value from this `partial_cmp` while calling `cstep` in `src/solver/linesearch/morethuente.rs`.

This panic comes about when comparing a NaN to an f64 which is undefined and panics. We can handle this case with iter::fold which takes the maximum value that isn't NaN, removing the need for unwraps and makes a panic while using the crate less likely.


stefan-k:
Thank you for your contribution! 

Did you by any chance find out the origin of the `NaN`s? I'm afraid that they may actually be the result of an underlying bug.

Are you by any chance able to provide a test-case where the panic occurs so that I can have a look at what is happening?

MattBurn:
The `NaN`s in my case occurred due to an overflow in the calculation of the determinant of a likelihood matrix, luckily I only needed the log determinant which I could calculate directly from the LU bypassing the overflow bug.

Before I came across the solution to the `NaN` I put in this fix to argmin. Unfortunately this bug was hard to reproduce as it is part of a machine learning application that is trying to find a global minimum in a non convex search space. The way I am attempting to do this is with many random restarts where very infrequently I would come across this panic which would crash out my application.

Now as I am restarting, a single restart crashing is not a problem and I can ignore it using Rust's error propagation but the panic was randomly preventing my application from finishing. 

The best solution would be something akin to:
`let s = tmp.iter().max_by(|a, b| a.partial_cmp(b)?)?;`

Where the function returns an error if a `NaN` is encountered which can then be handled by the Result from `next_iter`. I don't know of a way to achieve this as `max_by` cannot accept an Error

stefan-k:
Thanks a lot for the clarification! I will accept this PR and think about a way to deal with `NaN`s, maybe by just brute-force checking them before finding the maximum.

Issue #10, opened 2019-07-17
Benchmark L-BFGS


rth:
Thanks for this very nice package!

On the general topic of benchmarks, I would be quite interested to know how the L-BFGS implementation in argmin would compare to,
 - [liblbfgs](http://chokkan.org/software/liblbfgs/)
 - the [original Fortran implementation](http://users.iems.northwestern.edu/~nocedal/lbfgsb.html)

There are Python wrappers for these, in [PyLBFGS](https://pypi.org/project/PyLBFGS/) and [scipy.optimize](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin_l_bfgs_b.html#scipy.optimize.fmin_l_bfgs_b). So using [pyargmin](https://github.com/argmin-rs/pyargmin) it should be relatively easy to compare them (albeit with the Python overhead).

L-BGFS is very commonly used in machine learning (also see the ["Rust needs BFGS"](https://paulkernfeld.com/2018/08/06/rust-needs-bfgs.html) blog by the author of the [bfgs](https://github.com/paulkernfeld/bfgs) crate). I'm mostly interested in seeing of it's possible to get a comparable or faster implementation in Rust than the one in scipy.optimize, that would be more readable (and support both `f64` and `f32`).

Opening this issue mostly to report progress on it. If you have any suggestions (e.g. on the benchmark cases to use) I would be interested as well.

stefan-k:
Such a benchmark would be really interesting. I haven't compared against these implementations at all, so besides the performance differences, it may be possible that they do not produce the exact same results. 
In terms of performance, I suspect (or better, hope) that most time is spent on linear algebra. I think [flamegraph](https://github.com/ferrous-systems/flamegraph) could be a very useful tool for this.
I will see if I can find problems for the benchmark (which is something I wanted to do for a long time anyway).

nilgoyette:
I tested this today. I'm using the [original Fortran implementation v3](http://users.iems.northwestern.edu/~nocedal/lbfgsb.html). My application runs in

- **15s** with the L-BFGS-B version
- **18s** with the L-BFGS version

[...]

PR #49, opened 2020-04-14
Golden-section search

nilgoyette:
This is my first version of `Golden-section search`. I tested in our enterprise project and it's as good as my hardcoded version, but a little slower.

I know that there's no test, example, bench, nothing. I just want you to tell me what I did wrong. My own critic is that there's no way to choose the precision. If I understand correctly, everything is forced to f64 with `O: ArgminOp<Param = f64, Output = f64>,` so the user can't actually choose with

    type Param = f32/f64;
    type Output = f32/f64;

### DIFF
--- a/src/solver/goldensectionsearch/mod.rs
+++ b/src/solver/goldensectionsearch/mod.rs
@@ -106,18 +106,23 @@ where
             self.x2 = x2;
             self.f1 = op.apply(&self.x1)?;
             self.f2 = op.apply(&self.x2)?;
-            Ok(Some(ArgminIterData::new()))
+            if self.f1 < self.f2 {
+                Ok(Some(ArgminIterData::new().param(self.x1).cost(self.f1)))
+            } else {
+                Ok(Some(ArgminIterData::new().param(self.x2).cost(self.f2)))
+            }
         }
     }
 
     fn next_iter(
         &mut self,
         op: &mut OpWrapper<O>,
-        _state: &IterState<O>,
+        state: &IterState<O>,
     ) -> Result<ArgminIterData<O>, Error> {
         if self.tolerance * (self.x1.abs() + self.x2.abs()) >= (self.x3 - self.x0).abs() {
             return Ok(ArgminIterData::new()
-                .param(if self.f1 < self.f2 { self.x1 } else { self.x2 })
+                .param(state.param)
+                .cost(state.cost)
                 .termination_reason(TerminationReason::TargetToleranceReached));
         }
 
@@ -134,6 +139,10 @@ where
             self.f2 = self.f1;
             self.f1 = op.apply(&self.x1)?;
         }
-        Ok(ArgminIterData::new())
+        if self.f1 < self.f2 {
+            Ok(ArgminIterData::new().param(self.x1).cost(self.f1))
+        } else {
+            Ok(ArgminIterData::new().param(self.x2).cost(self.f2))
+        }
     }
 }
--- a/src/solver/goldensectionsearch/mod.rs
+++ b/src/solver/goldensectionsearch/mod.rs
@@ -35,51 +35,55 @@ const G2: f64 = 1.0 - G1;
 ///
 /// [Wikipedia](https://en.wikipedia.org/wiki/Golden-section_search)
 #[derive(Clone, Serialize, Deserialize)]
-pub struct GoldenSectionSearch<O: ArgminOp> {
-    min_bound: f64,
-    max_bound: f64,
-    init_estimate: f64,
-    tolerance: f64,
+pub struct GoldenSectionSearch<F> {
+    g1: F,
+    g2: F,
+    min_bound: F,
+    max_bound: F,
+    init_estimate: F,
+    tolerance: F,
 
-    x0: f64,
-    x1: f64,
-    x2: f64,
-    x3: f64,
-    f1: O::Output,
-    f2: O::Output,
+    x0: F,
+    x1: F,
+    x2: F,
+    x3: F,
+    f1: F,
+    f2: F,
 }
 
-impl<O> GoldenSectionSearch<O>
+impl<F> GoldenSectionSearch<F>
 where
-    O: ArgminOp,
-    O::Output: ArgminZero,
+    F: ArgminFloat,
 {
     /// Constructor
-    pub fn new(min_bound: f64, max_bound: f64) -> Self {
+    pub fn new(min_bound: F, max_bound: F) -> Self {
         GoldenSectionSearch {
+            g1: F::from(G1).unwrap(),
+            g2: F::from(G2).unwrap(),
             min_bound,
             max_bound,
-            init_estimate: 0.0,
-            tolerance: 0.01,
+            init_estimate: F::zero(),
+            tolerance: F::from(0.01).unwrap(),
             x0: min_bound,
-            x1: 0.0,
-            x2: 0.0,
+            x1: F::zero(),
+            x2: F::zero(),
             x3: max_bound,
-            f1: O::Output::zero(),
-            f2: O::Output::zero(),
+            f1: F::zero(),
+            f2: F::zero(),
         }
     }
 
     /// Set tolerance
-    pub fn tolerance(mut self, tol: f64) -> Self {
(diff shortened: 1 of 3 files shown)


src/solver/goldensectionsearch/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+            self.f1 = self.f2;
+            self.f2 = op.apply(&self.x2)?;
+        } else {
+            self.x3 = self.x2;
+            self.x2 = self.x1;
+            self.x1 = G1 * self.x2 + G2 * self.x0;
+            self.f2 = self.f1;
+            self.f1 = op.apply(&self.x1)?;
+        }
+        Ok(ArgminIterData::new())
```
stefan-k, line 137: If I understand this correctly, the smaller one of `self.f1` and `self.f2` is your current best cost function value and `self.x1` or `self.x2` is the corresponding parameter vector. If you could add them here to the `ArgminIterData` struct as well (via `param()` and `cost()`), then they will be available to observers. 

The same applies to line 119 regarding `cost()`.

You can then access these values in the `state` variable passed to `next_iter` via `get_param()` and `get_cost()`, which would spare you the check in line 120. But this is a matter of taste.

src/solver/goldensectionsearch/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub fn tolerance(mut self, tol: f64) -> Self {
+        self.tolerance = tol;
+        self
+    }
+}
+
+impl<O> Solver<O> for GoldenSectionSearch<O>
+where
+    O: ArgminOp<Param = f64, Output = f64>,
+{
```
stefan-k, line 83: It would be interesting to see if it is possible to solve the `f32`/`f64` problem with the `Float` trait of the crate `num_traits`. 

Something like this:

```rust
impl<O, F> Solver<O> for GoldenSectionSearch<O>
where
    O: ArgminOp<Param = F, Output = F>,
    F: Float,
{
    // ...
}
```

Or, maybe even easier (and more related to your initial question in the issue):

```rust
impl<O, F> Solver<O> for GoldenSectionSearch<O>
where
    O: ArgminOp<Param = F, Output = F>,
    F: PartialEq,
{
    // ...
}
```
(You will probably need to add more trait bounds.)

Both potential solutions may however require you to also be more generic in the definition of `GoldenSectionSearch`.

nilgoyette:
Our program runs GSS two times, they take around 26ms and 23ms. When I use this version, it's around 31ms and 29ms. Nothing terrible, but maybe just annoying that the same algo is suddenly 20% slower. I didn't activate any special features, no logging, no observer, no key-value, nothing. I don't understand what's taking 5-6ms more under the hood.

> I appreciate any feedback regarding the ergonomics when implementing a solver

I had difficulty implementing this simple algo. **Mostly because of the types.** At first I thought there would be a single type F for precision and that's it. Then I saw `O::Param` and `O::Output`, and almost only `f64` looking at your own code. At that time, I was really confused. And I saw no way to cast from one type to the other so I was even more confused. My other concerns were discussed in the linked issue.

As a user (and maybe contributor) of this lib, what matter is the speed, so disabling the stuff I don't need and choosing the float precision. I port a lot of Python and C++ code to Rust and it's sometime optimization algorithm, so we may have other methods to contribute, or maybe try to optimize those you already coded. For example, there's a lasso algorithm, ported from [SPAMS](http://spams-devel.gforge.inria.fr/). We don't want it in our codebase, but it probably doesn't fit here? I must warn you though, I'm not a mathematician at all. I sometime don't even understand what I'm porting, but the results are identical or really close and it's often faster. That's all we can hope for :)

src/solver/goldensectionsearch/mod.rs
nilgoyette: I tried this and it went well, once I figured this line

    F: Float + Serialize + DeserializeOwned,

Everything was now in the type choosed by the user. I had one problem though. `ArgminIterData::cost` is defined as a `f64`, so I can't write

    ArgminIterData::new().param(self.x1).cost(self.f1)
    expected `f64`, found type parameter `F`

I would need to cast it and that's exactly what I don't want to do.

stefan-k:
> Our program runs GSS two times, they take around 26ms and 23ms. When I use this version, it's around 31ms and 29ms. Nothing terrible, but maybe just annoying that the same algo is suddenly 20% slower. I didn't activate any special features, no logging, no observer, no key-value, nothing. I don't understand what's taking 5-6ms more under the hood.

There is quite some stuff going on under the hood in the [executor](https://github.com/argmin-rs/argmin/blob/master/src/core/executor.rs), I suppose that might be the reason for the 5-6ms. If it stays at 5-6ms for problems of any complexity and for any solver, than I guess that's fine, but we've seen in #10 that it may be worse. Performance is certainly an issue which needs to be addressed.

> I had difficulty implementing this simple algo. Mostly because of the types. 

Being generic over all kinds of types is difficult and I'm aware that ergonomics suffers from it (particularly for a developer of solvers). I could have just supported `ndarray`, but in my own research, I often had to deal with "special" parameter vectors which required certain operations on them to be implemented manually. I didn't want parameter vectors to be limited to vectors of floats.
But that's independent of the problem with `f32` and `f64` where I totally agree that it should be generic. I will address this in more detail in the comment above. 

> As a user (and maybe contributor) of this lib, what matter is the speed, so disabling the stuff I don't need and choosing the float precision. 

I think we need to start to excessively benchmark the code and identify the time consuming parts. I'm sure there are lots of places where things can be improved, and in cases where it is not possible, we may be able to put those things behind feature gates.

> I port a lot of Python and C++ code to Rust and it's sometime optimization algorithm, so we may have other methods to contribute, or maybe try to optimize those you already coded. For example, there's a lasso algorithm, ported from SPAMS. We don't want it in our codebase, but it probably doesn't fit here? 

I don't see why it shouldn't fit here, unless there are licensing issues. I really appreciate every contribution, particularly because it creates feedback like this which I desperately need to improve the library.

> I must warn you though, I'm not a mathematician at all. I sometime don't even understand what I'm porting, but the results are identical or really close and it's often faster. That's all we can hope for :)

I'm not a mathematician either and this is also how I work, so there is no need to worry ;)

src/solver/goldensectionsearch/mod.rs
stefan-k: Thanks for giving this a try! I'm happy to see that it almost works. I guess now is the time to make `ArgminIterData` generic over `f32` and `f64`. I will try to get at this on the weekend, I hope that is not too late for you.

stefan-k:
With PR #50 it should be possible to choose the float precision. Ergonomics definitely suffers from it (for developers of solvers), because one can't just use `1.0`, but instead has to write `F::from_f64(1.0).unwrap()`. However, I don't think this has a performance impact, because the compiler should optimize it all away.

stefan-k:
PR #50 is merged and you could be the first to give it a try ;)

src/solver/goldensectionsearch/mod.rs
nilgoyette: Thanks to you and this huge float branch, I was able to test it properly.

nilgoyette:
I gave it a try. Imo, this is a more sensible design for simple methods like Brent and GSS. I like it.

src/solver/goldensectionsearch/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub fn tolerance(mut self, tol: F) -> Self {
+        self.tolerance = tol;
+        self
+    }
+}
+
+impl<O, F> Solver<O> for GoldenSectionSearch<F>
+where
+    O: ArgminOp<Output = F, Param = F, Float = F>,
+    F: ArgminFloat + std::cmp::PartialOrd,
```
stefan-k, line 86: I think `Float` already implies `PartialOrd`: https://rust-num.github.io/num/num/trait.Float.html

nilgoyette: I probably added this trait before `ArgminFloat`. I removed it.

stefan-k:
Thanks again for your contribution and the feedback! :)

PR #50, opened 2020-04-18
Generic floats

stefan-k:
With this PR, the user can choose the floating point precision.

TODOs:
* [X] almost everything
* [X] Fix deserialization of `Particeswarm`
* [X] Fix `modify` in `ArgminOp`

At the end, I realized that it may be smarter to carry the type around in the `ArgminOp` (as an associated type). I'll have to think about it. 

* [X] Make precision part of `ArgminOp`


rth:
Ability to run with f32 is definitely important. Thanks for working on this !

Issue #42, opened 2020-04-08
Error handling and dealing with `NaN`s in `MoreThuenteLineSearch`


stefan-k:
This is related to #38. 

Prior to #38, `NaN`s (wherever they may come from) would cause `MoreThuenteLineSearch` to panic. Since #38, `NaN`s will be ignored. The appropriate way to handle such a case would be to check for `NaN`s and throw an error, so that this case can be handled by the user.

stefan-k:
Fixed in #57. Thanks @MattBurn !

PR #57, opened 2020-04-27
MoreThuente: Added error check for NaN or Inf

MattBurn:
Added a check for NaN or Inf in the cstep calculation of the MoreThuente linesearch using iter any. Returns an error if an invalid value is encountered.

Should resolve issue #42


stefan-k:
Thanks a lot for this contribution! :) Looks good to me!

PR #68, opened 2020-08-14
Add nalgebra support

Maher4Ever:
This PR adds support for the [`nalgebra`](https://nalgebra.org) crate. This allows using static matrices with `argmin`. It also means that operations that panicked before now result in a compilation error (e.g., multiplying incompatible static matrices). 

Not all math traits were implemented; only the ones that were implemented for `ndarray` are now implemented for `nalgebra`. A usage example was also added for the Gauss-Newton solver.

The implementations of `argmin`'s math traits use a lof of generic code to support all the matrices provided by `nalgebra` (both static and dynamic ones). However, the disadvantage is that any change in `nalgabra`'s definitions of matrices or the functions they support will break the implementations in `argmin`.

NB: Thank you for providing and maintaining such a great optimization crate with support for many solvers. It saved us a lot of time :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ bincode = "1.1.4"
 ctrlc = { version = "3.1.2", optional = true }
 gnuplot = { version = "0.0.37", optional = true}
 paste = "1.0.0"
+nalgebra = { version = "0.21.1", optional = true, features = ["serde-serialize"] }
 ndarray = { version = "0.13", optional = true, features = ["serde-1"] }
 ndarray-linalg = { version = "0.12", optional = true }
 ndarray-rand = {version = "0.11.0", optional = true }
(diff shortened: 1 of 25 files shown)


stefan-k:
Wow! This is amazing, thank you for this massive contribution! 

> However, the disadvantage is that any change in `nalgabra`'s definitions of matrices or the functions they support will break the implementations in `argmin`.

That's probably a general problem of this approach. I'm open to any ideas that may improve this situation. 
I've designed the math traits with `ndarray` and `Vec` in mind, which in some situations may be suboptimal for `nalgebra`. If you have suggestions for improvements please let me know. 

I'm not sure, but I think the CI is failing because `finitediff` does not support `nalgebra` (yet). 

I will try to give this a thorough review in the next couple of days.

Thanks again! :)

Maher4Ever:
@stefan-k You're more than welcome :)

The CI is failing because I accidentally committed an example I was working on that (indeed) isn't compiling due to `finitediff` not supporting `nalgebra`. I removed that example from the branch to solve the issue. Hopefully, the checks will succeed now.

stefan-k (approved):
Thanks again for your contribution! :)

PR #77, opened 2020-11-01
Brent's optimization method

### LINKED ISSUE #76: Implement Brent minimization method
Hello,

Unless I am mistaken, Brent method is the name for two algorithms: [one for root-finding](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.brentq.html) and [one for minimization](https://docs.scipy.org/doc/scipy/reference/optimize.minimize_scalar-brent.html). They are respectively described under the names `zero` and `localmin` in chapters 4 and 5 of the book _Algorithms for Minimization without Derivatives_ by Brent (1973). Although they share the same ideas, the first is not directly applicable to the second problem unless we know the derivative of the function to optimize.

It seems that the function implemented in `argmin` is Brent's root-finding method (currently the only one described on Wikipedia). Hence I think that we are still missing Brent's minimization method. I would like to implement the minimization method, but my question is what to do of the root-finding method already present? Is root-finding one of the purposes of `argmin`? In which case, maybe would it be appropriate to separate root-finding and optimization algorithms, since those are two different kinds of problems?

cc @xemwebe

Armavica:
Fixes #76.


Armavica:
Not necessarily to be merged immediately, just to show Brent's optimization method and discuss its place within `argmin`.

Armavica:
Another question that I have is what to do with algorithms that do not need an initial guess (like this one). Actually it can use an initial guess, but by default it computes its own. Should we pass `F::nan()` (for no initial guess) or an actual number (when we have a guess) to the executor, or should we add a method to the solver builder `set_init`, and ignore the value of the executor?

stefan-k:
Thank you for your contribution! Unfortunately I don't have time right now, but I'll hopefully be able to have a look by the end of the week. 

@xemwebe If possible, I'd also really appreciate your feedback on this :)

xemwebe:
@stefan-k I am not sure what kind of feedback you expect from me, other than the comments I added to the Issues thread. I wasn't aware that there exist two methods which could be claimed to be Brent's. There might be a reason why the Brent's minimization method is less well known than his rout finding method. I don't know.

xemwebe:
> 
> 
> Another question that I have is what to do with algorithms that do not need an initial guess (like this one). Actually it can use an initial guess, but by default it computes its own. Should we pass `F::nan()` (for no initial guess) or an actual number (when we have a guess) to the executor, or should we add a method to the solver builder `set_init`, and ignore the value of the executor?

The rust way would probably to pass the guess as an Option<guess>, instead of using magic numbers (like nan), but if this would require changing the interface, which is not a good idea if its only for rare exceptions to the rule. Personally, I probably would prefer to have to constructors. Or maybe make the function calculating the initial guess public and let the user decide whether she wants to use it or not.

stefan-k:
As suggested by @xemwebe I'd personally probably go for two constructors. However it may not be the most logical thing to do, because one could argue that the initial value is not necessarily a part of the solver, but only gets its meaning when combined with an instance of an optimization problem. Therefore it would make the most sense if it would be possible to set it via the executor (or as part of the problem definition). But right now I don't see a way to do this in a consistent manner, therefore I would go for two constructors or a builder pattern.

Armavica:
Two constructors for the solver or for the executor? Would there be an issue with having `Executor::new(problem, solver)` and `Executor::new_with_x0(problem, solver, x0)`? Or even `Executor::new(problem, solver).set_initial(x0)`?

stefan-k:
I'm sorry, I seem to have not read your post properly and therefore I misunderstood the problem quite a bit. As I see it, I don't think it is straightforward to allow for not providing an initial guess to the executor. It is not impossible, but it requires changes in the internals of the `Executor`, in particular in `IterState`. While I'm also not a huge fan of magic numbers, I think in this case it might be reasonable to use `Nan` to indicate no initial parameter. I also like the suggestion by @xemwebe to make the function which computes the initial guess public, however I don't know if this is feasible in your case. 

Would you be interested to have a look at the handling of initial values in the `Executor`? If not, I will open an issue to keep this in mind for the future. I believe that the `param*` fields of `IterState` need to be handled similarly to the `grad*` fields.

stefan-k:
Apologies that this PR has not gotten any of my attention for a while! I'm still very interested in this contribution and would be very happy if we could revive it.
`argmin` has changed a lot in the meantime. In particular, since #198, it is not necessary anymore to provide an initial guess! :)

Armavica:
Thank you for reminding me of this issue! I will try to take care of it this week.

Armavica:
Indeed a lot of things changed, congrats for the large and very nice update!
I updated the PR, but I am not sure if I did everything correctly, because I am not very familiar with the new API yet.
If I remember correctly, the only remaining question was the naming of this optimization method, with respect to the `brent` root-finding method which already exists. You suggested sharing the namespace `solvers::brent` in a comment on the related issue, in that case would `solvers::brent::root` and `solvers::brent::opt` be adequate?

stefan-k:
Thanks for the quick update and thanks for the feedback on the recent changes! If you experienced any strangeness due to the (new) design please let me know. 

I've had a look at your implementation and couldn't find anything to comment on. 

Regarding the naming: I'm somewhat leaning towards `solvers::brent::BrentRoot` and `solvers::brent::BrentOpt`. I would still have them in separate files (`opt.rs` and `root.rs`) but reexport them on the `solvers::brent` level. 
I'm not a huge fan of the `Opt` in `BrentOpt` but it is the most explicit way of separating them and it avoids (a bit of) confusion when users update the library. I don't have a very strong opinion about it and I'm of course open for other suggestions/arguments.

Once we've decided on the names I think this is ready to go! :)

Armavica:
I did like you asked, it seems reasonable to me.

argmin/src/solver/brent/brentopt.rs
```diff
@@ -0,0 +1,292 @@
...
+    pub fn set_tolerance(mut self, eps: F, t: F) -> Self {
+        self.eps = eps;
+        self.t = t;
+        self
+    }
+}
+
+impl<O, F> Solver<O, IterState<F, (), (), (), F>> for BrentOpt<F>
+where
+    O: Operator<Param = F, Output = F>,
```
stefan-k, line 111: I just noticed that you require `O` to implement `Operator` here (probably because `BrentRoot` does the same). I was wondering if `CostFunction` would make more sense in this context for both Brent methods (even though it essentially boils down to the same thing because both Brent methods optimize a single number). I got this idea because I noticed that you store the result of `apply` as cost function value in `state`.

stefan-k (changes requested):
Thanks, this looks really good! There's only one small comment that needs clarification.

argmin/src/solver/brent/brentopt.rs
Armavica: You are right, I took this from `BrentRoot` without really asking myself why. However I have to say that I am not sure of the difference between `Operator` and `CostFunction`, they look very similar to me… Would you mind elaborating on when to use one rather than the other? I think that I see a pattern in the code where `CostFunction`'s `Output` is intended to be scalar whereas `Operator`'s `Output` not necessarily, is that correct?

stefan-k: I must have changed `ArgminOp` to `Operator` in `BrentRoot` without giving it much thought. `Operator` is meant for cases where one would for instance need to apply a matrix to a vector (such as in `ConjugateGradient`). The returned type would then be some sort of vector. `CostFunction` on the other hand always returns a scalar. I'm currently in the process of rewriting the documentation and I definitely need to make this more clear.

Armavica: I see, thanks! I replaced `Operator` with `CostFunction` in both cases.

stefan-k (approved):

stefan-k:
Good to go! Thanks a lot again for this contribution and your patience! :)

Issue #2, opened 2019-02-10
Bug in HagerZhangLineSearch


stefan-k:
There seems to be a bug in HagerZhangLineSearch (for instance, running the steepestdescent or nonlinear_cg examples with HagerZhangLineSearch will not converge towards the solution).

whereistejas:
Hi @stefan-k,
I'm interested in this particular issue and would like to help fix it (if it is still relevant). Could you give me some pointers on where to get started? I'm also interested in contributing in general, so please feel to direct to me more urgent or important issues. :)

stefan-k:
Hi @whereistejas ,

thanks for your interest to contribute! It's been a while since I opened this issue, so I can't remember the details anymore. The implementation follows this paper [1]. The code can be found in [2]. I think a good way to move forward is to reproduce the errors by running the [`steepestdescent`](https://github.com/argmin-rs/argmin/blob/master/examples/steepestdescent.rs) and [`nonlinear_cg`](https://github.com/argmin-rs/argmin/blob/master/examples/nonlinear_cg.rs) examples with the `HagerZhangLineSearch`. I suspect a wrong `<`, `>`, `<=` or `>=`. 

While going through the implementation, I noticed a comment that not all stopping criteria are implemented, but I don't think that this is causing the problem. 

If you have any questions feel free to get in touch via the issues or via gitter. I'm happy to help, but I may unfortunately be a bit slow to respond currently!
 
[1] William W. Hager and Hongchao Zhang. "A new conjugate gradient method with guaranteed descent and an efficient line search." SIAM J. Optim. 16(1), 2006, 170-192.
[2] https://github.com/argmin-rs/argmin/blob/master/src/solver/linesearch/hagerzhang.rs

whereistejas:
Hi @stefan-k ,
I compared the output of the two line search methods in [`steepestdescent`](https://github.com/argmin-rs/argmin/blob/master/examples/steepestdescent.rs). 

The output for Hager-Zhang method is: 
```
[
  1.1999999999997424,
  1.2000000000001065
]
```
Where as the output for More-Thuente method is:
```
[
  1.0724442427550391,
  1.151531565672518
]
``` 
So, I'm guessing, this is discrepancy you are talking about?

stefan-k:
[...]

PR #92, opened 2021-01-21
WIP: Corrections in Hager-Zhang line search method

whereistejas:
Changes being introduced in PR:
1. corrections to update steps U1, U2 and U3
2. different example in examples/hagerzhang.rs

Fixes issue: #2

### DIFF
--- a/examples/hagerzhang.rs
+++ b/examples/hagerzhang.rs
@@ -11,6 +11,7 @@ use argmin::solver::linesearch::HagerZhangLineSearch;
 
 struct problem {}
 
+// cost_function
 // f(x)=(x−3) x^{3} (x−6)^{4}
 pub fn cf(params: &Vec<f64>) -> f64 {
     let x = params[0];
@@ -19,6 +20,7 @@ pub fn cf(params: &Vec<f64>) -> f64 {
     value
 }
 
+// derivative of cost_function
 // df(x) = (x - 6)**5 *
 // (x**3*(x - 6)*(x - 3)**(x**4)*(x + 4*(x - 3)*log(x - 3)) + 6*(x - 3)**(x**4 + 1)) /
 // (x - 3)
@@ -72,9 +74,6 @@ fn run() -> Result<(), Error> {
     let init_cost = operator.apply(&init_param)?;
     let init_grad = operator.gradient(&init_param)?;
 
-    println!("init_cost: {:?}", init_cost);
-    println!("init_grad: {:?}", init_grad);
-
     // Run solver
     let res = Executor::new(operator, solver, init_param)
         .add_observer(ArgminSlogLogger::term(), ObserverMode::Always)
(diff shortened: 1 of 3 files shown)


examples/hagerzhang.rs
```diff
@@ -6,35 +6,58 @@
 // copied, modified, or distributed except according to those terms.
 
 extern crate argmin;
-extern crate argmin_testfunctions;
 use argmin::prelude::*;
 use argmin::solver::linesearch::HagerZhangLineSearch;
-use argmin_testfunctions::{sphere, sphere_derivative};
 
-struct Sphere {}
+struct problem {}
```
stefan-k, line 12: Struct names should start with a capital letter.

src/solver/linesearch/hagerzhang.rs
```diff
@@ -325,17 +325,17 @@ where
             self.update(op, (a_x, a_f, a_g), (b_x, b_f, b_g), (c_x, c_f, c_g))?;
 
         // S2
-        if (c_x - bb_x).abs() < F::epsilon() {
+        if c_x == bb_x {
```
stefan-k, line 328: `c_x`, `bb_x`, `aa_x`, and so on are floating point numbers. Floating point numbers should never be compared via `==`, because due to the limited precision, rounding errors will accumulate differently for different mathematical operations. Therefore, even if two numbers should mathematically be the same, they could be different when represented as floating point numbers. Therefore the general approach is to check that the difference between two numbers is small.

src/solver/linesearch/hagerzhang.rs
```diff
@@ -325,17 +325,17 @@ where
...
             c_bar_x = self.secant(b_x, b_g, bb_x, bb_g);
         }
 
         // S3
-        if (c_x - aa_x).abs() < F::epsilon() {
+        if c_x == aa_x {
```
stefan-k, line 333: Same here.

src/solver/linesearch/hagerzhang.rs
```diff
@@ -325,17 +325,17 @@ where
...
             c_bar_x = self.secant(a_x, a_g, aa_x, aa_g);
         }
 
         // S4
-        if (c_x - aa_x).abs() < F::epsilon() || (c_x - bb_x).abs() < F::epsilon() {
+        if c_x == bb_x && c_x == aa_x {
```
stefan-k, line 338: Same here.

stefan-k (changes requested):
Thank you for your PR! Unfortunately, I still see the exact same issue with your fix when running the `steepestdescent` example:

```
ArgminResult:
    param (best):  [1.1999999999997424, 1.2000000000001065]
```

The changes you made are unlikely to improve the situation as explained in my comments.

stefan-k:
Closing due to inactivity. Feel free to reopen if you decide to continue working on it.

PR #101, opened 2021-03-22
Fix lazy evaluation of gradients in line searches

w1th0utnam3:
The Backtracking and Hager Zhang line searches always evaluated the gradient, even if it was already present (because of usage of `unwrap_or`). In addition, More Thuente would panic if the gradient computation returns an error (because of usage of `unwrap`). This PR fixes this.

Unrelated: With respect to your usage of dependabot I would recommend using squash & merge in the Github UI to avoid littering your commit history with tons of merge commits. But feel free to ignore this opinion.

### DIFF
--- a/src/solver/linesearch/backtracking.rs
+++ b/src/solver/linesearch/backtracking.rs
@@ -121,7 +121,10 @@ where
             cost
         };
 
-        self.init_grad = state.get_grad().unwrap_or(op.gradient(&self.init_param)?);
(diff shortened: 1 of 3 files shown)


stefan-k:
Good catch, thanks a lot for this PR! 

Regarding merge/squash: I have to admit that I just didn't pay that much attention, but I agree that squash & merge would be more appropriate. Thanks for this comment!

Issue #111, opened 2021-04-29
Performance Enhancements


sdd:
First of all, I'd like to thank everyone involved in making this crate. I've been using it in multiple places in a project of mine for the past few months, and I love the ergonomics and find the crate a joy to use - Kudos!

Anyway, I was doing some performance tuning on my code recently and discovered that in my calls to `Executor::run`, only about 44% of processor time was spent inside my `ArgminOp::apply`. When looking at where the rest of the time was being spent, I noticed three main areas:

1. `alloc::fmt::format`, `ArgminKV::merge`, and `drop ArgminKV`. This would be expected if I was using an observer, but this overhead happens even with no registered observers.
2. `Instant::elapsed` and `Instant::now`. These are called even if I don't care about the elapsed time (as is often the case once I've finished tuning the optimisation params)`.
3. `ArgminCheckpoint::store_cond`. This allocates a string, concats it with another string, and deallocs the result on exit - every single time through the `Executor::run` loop, even if Checkpoints are off, or this loop is not a checkpointed loop.

Here's a section of my flamechart from cargo-flamechart to illustrate this.
Green boxes highlight my code inside the apply function, magenta is for point (1) above, red for point (2), and blue for point (3)

![image](https://user-images.githubusercontent.com/323439/116487949-1cacba80-a889-11eb-9c70-da286b835746.png)

I've got a branch in my fork where I've addressed points (1) and (3) by some refactoring, which I don't think will cause any issues (but you all know better than me, I may be wrong!). I've also addressed (2) by introducing a config flag to make the timer optional, similarly to how enabling/disabling Ctrl-C works. You can see my changes here: https://github.com/sdd/argmin/commit/a67f56a1313aea7dfbd7f14459942bbd7ae2182b

With the modifications above in place, I get a 44% speedup, and I can see that the executor now spends 93% of its time running my `ArgminOp::apply`, rather than 44% of the time. My cargo flamechart now looks more like this:

![image](https://user-images.githubusercontent.com/323439/116487539-1d911c80-a888-11eb-88f1-60596767e095.png)

I realize that my workload may be atypical to many others, as it might be quite light. I'm really just doing some matrix and trig operations. But, even with intensive `apply` functions, some performace gains are there to be had for others.

I'm happy to raise a PR and whip it into the kind of shape you'd expect it to be in if you think this is a worthwhile change.

Thanks!

stefan-k:
[...]

PR #112, opened 2021-04-29
Performance Enhancements

sdd:
* eliminate calls to ArgminKV when there are no observers
* make the timer optional
* only create/append/destroy filename string when an iteration will write a checkpoint

See https://github.com/argmin-rs/argmin/issues/111 for more detail.

### DIFF
--- a/src/core/serialization.rs
+++ b/src/core/serialization.rs
@@ -49,6 +49,7 @@ pub struct ArgminCheckpoint {
     mode: CheckpointMode,
     directory: String,
     name: String,
+    filename: String,
 }
 
 impl Default for ArgminCheckpoint {
@@ -57,6 +58,7 @@ impl Default for ArgminCheckpoint {
             mode: CheckpointMode::Never,
             directory: ".checkpoints".to_string(),
             name: "default".to_string(),
+            filename: "default.arg".to_string(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/core/serialization.rs
```diff
@@ -103,6 +103,14 @@ impl ArgminCheckpoint {
         self.name.clone()
     }
 
+    /// Get filename for checkpoint
+    #[inline]
+    fn filename(&self) -> String {
+        let mut filename = self.name();
+        filename.push_str(".arg");
```
stefan-k, line 110: Maybe it would be even better to store `filename` in the `ArgminCheckpoint` struct and have it set to `name` + `".arg"` when when `set_name` is called (or to `default.arg` when `default()` is called). This method would then just return `self.filename`. That way one would also profit from your performance improvements when using the checkpointing mechanism.

stefan-k (changes requested):
Thanks again! I only have one minor suggestion.

src/core/serialization.rs
sdd: I was considering this, but didn't, in case a consumer called `set_name()` in the meantime. On reflection, I could simply update the filename inside of `set_name()` to mitigate that. Will update 👍🏼

stefan-k (approved):

PR #118, opened 2021-06-01
Squash warnings for Nalgebra 0.26.x

CattleProdigy:
- Switched to `OMatrix` for the the NAlgebra implementation of the math traits. This will work with both static and dynamic matrices. I confirmed that these work even with `SMatrix`.
- Bumped NAlgebra to the latest released version


stefan-k:
Thanks a lot, I really appreciate it! There is nothing to complain about :)

CattleProdigy:
Thanks!

PR #136, opened 2021-09-07
Make use of `slog` a feature

ThatGeoGuy:
See the two commits provided for more context. 

The main crux of this is that I want to use `argmin`, but I don't want to include `slog` in the process. It makes sense then to make it a feature. Given how prolific this feature is throughout the library (it is the default terminal logger, even the file logger just goes to bincode / JSON), I opted to leave it as a default feature. Of course, if users don't want `slog` themselves, then they just have to specify the following in their `Cargo.toml`:

```TOML
[dependencies.argmin]
version = "0.4"
default-features = false
features = ["nalgebral", "ndarrayl"]
```

With whatever features they do want enabled explicitly in the `dependencies.argmin.features` array.
The first commit just cleaned up a bunch of clippy lints. The command I'm using is:

```bash
$ cargo clippy --all-targets --all-features -- -D warnings
```

I'm also on clippy (0.1.54). Hopefully the reasoning in the commit message tracks.

### DIFF
--- a/src/core/iterstate.rs
+++ b/src/core/iterstate.rs
@@ -368,11 +368,22 @@ mod tests {
         assert_eq!(state.get_prev_param(), param);
         assert_eq!(state.get_best_param(), param);
         assert_eq!(state.get_prev_best_param(), param);
-        assert_eq!(state.get_cost(), std::f64::INFINITY);
-        assert_eq!(state.get_prev_cost(), std::f64::INFINITY);
-        assert_eq!(state.get_best_cost(), std::f64::INFINITY);
-        assert_eq!(state.get_prev_best_cost(), std::f64::INFINITY);
-        assert_eq!(state.get_target_cost(), std::f64::NEG_INFINITY);
+
+        assert!(state.get_cost().is_infinite());
+        assert!(state.get_cost().is_sign_positive());
+
+        assert!(state.get_prev_cost().is_infinite());
+        assert!(state.get_prev_cost().is_sign_positive());
+
+        assert!(state.get_best_cost().is_infinite());
+        assert!(state.get_best_cost().is_sign_positive());
+
+        assert!(state.get_prev_best_cost().is_infinite());
+        assert!(state.get_prev_best_cost().is_sign_positive());
+
+        assert!(state.get_target_cost().is_infinite());
+        assert!(state.get_target_cost().is_sign_negative());
+
         assert_eq!(state.get_grad(), None);
         assert_eq!(state.get_prev_grad(), None);
         assert_eq!(state.get_hessian(), None);
@@ -380,7 +391,9 @@ mod tests {
         assert_eq!(state.get_jacobian(), None);
         assert_eq!(state.get_prev_jacobian(), None);
         assert_eq!(state.get_iter(), 0);
-        assert_eq!(state.is_best(), true);
+
+        assert!(state.is_best());
+
         assert_eq!(state.get_max_iters(), std::u64::MAX);
         assert_eq!(state.get_cost_func_count(), 0);
         assert_eq!(state.get_grad_func_count(), 0);
@@ -394,13 +407,15 @@ mod tests {
 
         state.cost(cost);
 
-        assert_eq!(state.get_cost(), cost);
-        assert_eq!(state.get_prev_cost(), std::f64::INFINITY);
+        assert_eq!(state.get_cost().to_ne_bytes(), cost.to_ne_bytes());
+        assert!(state.get_prev_cost().is_infinite());
+        assert!(state.get_prev_cost().is_sign_positive());
 
         state.best_cost(cost);
 
-        assert_eq!(state.get_best_cost(), cost);
-        assert_eq!(state.get_prev_best_cost(), std::f64::INFINITY);
+        assert_eq!(state.get_best_cost().to_ne_bytes(), cost.to_ne_bytes());
+        assert!(state.get_prev_best_cost().is_infinite());
+        assert!(state.get_prev_best_cost().is_sign_positive());
 
         let new_param = vec![2.0, 1.0];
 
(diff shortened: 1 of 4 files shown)


src/core/iterstate.rs
```diff
@@ -368,19 +369,23 @@ mod tests {
...
-        assert_eq!(state.get_prev_cost(), std::f64::INFINITY);
-        assert_eq!(state.get_best_cost(), std::f64::INFINITY);
-        assert_eq!(state.get_prev_best_cost(), std::f64::INFINITY);
-        assert_eq!(state.get_target_cost(), std::f64::NEG_INFINITY);
+
+        assert!(state.get_cost().is_infinite());
+        assert!(state.get_prev_cost().is_infinite());
+        assert!(state.get_best_cost().is_infinite());
+        assert!(state.get_prev_best_cost().is_infinite());
+        assert!(state.get_target_cost().is_infinite());
```
stefan-k, line 377: These tests make sure that the initial values are set correctly. Unfortunately, `.is_infinite()` does not consider the sign. For instance, it is important that `target_cost` is initialized with `NEG_INFINITY` and not `INFINITY`.

src/core/iterstate.rs
```diff
@@ -418,23 +423,23 @@ mod tests {
 
         state.cost(new_cost);
 
-        assert_eq!(state.get_cost(), new_cost);
-        assert_eq!(state.get_prev_cost(), cost);
+        assert_relative_eq!(state.get_cost(), new_cost, epsilon = f64::EPSILON);
```
stefan-k, line 426: My intention was to make sure that both numbers are exactly the same. I was not aware that such a comparison should be done using `.to_ne_bytes()`. Could you elaborate why this is preferable over comparing floats directly, please? I'm curious because I always thought that comparing floats directly boils down to bitwise comparison. 
However, for the sake of this test, `assert_relative_eq!` is fine too.

stefan-k (changes requested):
Thanks a lot for this PR! It makes absolutely sense to allow for disabling any `slog` related code. 
I only have a few minor comments regarding the test in `iterstate.rs`.

src/core/iterstate.rs
ThatGeoGuy: Is the sign relevant for all of them? It is possible to do direct byte checks here, but it would be somewhat uncommon to see that, which is why I opted for `is_finite()` checks. 

I can replace these with:

```rust
assert_eq!(state.get_target_cost().to_ne_bytes(), f64::NEG_INFINITY.to_ne_bytes());
```

Or positive infinity if that is relevant to the test itself.

src/core/iterstate.rs
ThatGeoGuy: Direct equality _would_ do a byte comparison on two numbers. If that's what is intended here though, it isn't clear (especially since a lot of argmin operations may choose any floating point type, even if this test is `f64`).

A common error in floating point comparisons is to compare two floats directly (e.g. `0.3f64` to `0.1f64`). However, this is almost exclusively wrong and conveys the wrong intent. Because of the way floats are represented, a computed value of `0.3` will not necessarily be "byte equal" to the literal `0.3`. 

So there's basically two ways you can compare these:

* Are they the exact same bytes?
* Are these numbers approximately equal (relative to some error margin)?

In the first case you want to compare the bytes, so the correct way to write the code would be:

```rust
assert_eq!(state.get_cost().to_ne_bytes(), new_cost.to_ne_bytes());
```

Using `to_ne_bytes` gives us two things:

1. It avoids the `clippy` lint in our tests (and I would argue you should lint your tests and examples the same way you lint your library, but maybe that's a personal preference). You probably didn't see this because you just called `cargo clippy`, which won't lint the tests or examples by default. You'll want `cargo clippy --all-features --all-targets -- -D warnings`. At the very least, `--all-targets` is needed here to get clippy to look through your tests.
2. It signals to the reader of the code that you're checking for byte-equality, not an approximation.

The second is the most important factor. If you expect a sentinel value (e.g. INFINITY, NEG_INFINITY, NAN) then byte comparisons are your best shot at that, and the code should communicate this. Case in point: I assumed these values could be approximate and would not return the exact bytes necessary, so I went for option two immediately.

**NOTE**: You could also compare `to_le_bytes` or `to_be_bytes`, but the native endian (ne) encoding is a better bet because it won't require any byte swizzling under the hood.

If you do want to check equality relative to some error epsilon, then the code I used above does what you're looking for. I am happy to change these if you think byte-equality is the proper check here. Is that the desired behaviour? 

**Another note about clippy**: Testing here, it seems that `cargo clippy --all-targets` is broken, due to missing some required features in the examples. It's possible to specify required features for each example but that makes the `Cargo.toml` somewhat noisy. I can also wrap that into this branch, but an easy way around it is just to use `--all-features` as well.

stefan-k: Thanks a lot for your detailed explanation! 

> You probably didn't see this because you just called cargo clippy, which won't lint the tests or examples by default.

You are absolutely right. :facepalm: I will adapt this in the CI soon.

> A common error in floating point comparisons is to compare two floats directly (e.g. 0.3f64 to 0.1f64). However, this is almost exclusively wrong and conveys the wrong intent. Because of the way floats are represented, a computed value of 0.3 will not necessarily be "byte equal" to the literal 0.3.

I am aware of that, but in this case the numbers are not the result of a computation but instead are set manually. 
I would prefer to check for byte-equality since the numbers should not change in any way. I absolutely agree with you though that the way I have done it does not show my intention. Therefore I would tend towards the `.to_ne_bytes()` solution, because like you said, it makes it clear that byte-equality is intended.

src/core/iterstate.rs
stefan-k: Yes, the sign is relevant for all of them. What do you think about asserting a combination of `.is_finite()` and `.is_sign_positive()` or `.is_sign_negative()`? I believe this would most clearly show what the intention is.

ThatGeoGuy: That's actually a fantastic idea, and preserves the intent. I should have checked to see a predicate for signage as well.

I will push a fixup commit up for this tomorrow morning.

src/core/iterstate.rs
ThatGeoGuy: Excellent. Given that, I'll push a fixup for this to do byte comparison tomorrow morning as well.

Cheers, and thanks for helping get to the final solution here.

src/core/iterstate.rs
ThatGeoGuy: See c163b362812fa96c2000502c293ce3d2823d4648. 

If the changes are good to go, let me know and I'll do an autosquash to wrap this back into the original commit.

src/core/iterstate.rs
ThatGeoGuy: See f783b9b09e31089651574c503a1bb298bc27bd07

Same as before, if the commit is good to go, I can autosquash and force-push before the merge.

stefan-k:
Thanks a lot for your contribution and for the fruitful discussion! :) Let me know when you think it is ready to go and I will merge the PR.

ThatGeoGuy:
Apologies, I saw this and completely forgot about it (oof).

I've force pushed, should be good to merge now.

stefan-k:
No problem! Thanks again! :)

stefan-k (approved):

PR #139, opened 2021-11-05
allow any RNG, improve SA example

### LINKED ISSUE #138: XorShiftRng has poor statistical quality
`rand_xorshift::XorShiftRng` uses Xorshift128 (also called `xor128`) which is from Marsaglia's original paper. It has significant weaknesses ([[0]](https://doi.org/10.1145%2F1113316.1113319)[[1]](http://pracrand.sourceforge.net/Tests_results.txt)). It was replaced as `rand`'s default noncryptographic PRNG in 2018.

Stochastic optimization requires high quality random numbers and Xorshift128 does not provide that.

argmin could use a sensible default like `rand`'s current `xoshiro256++` or even allow users to select one via traits (I'm willing to write a PR for that).

TheIronBorn:
Resolves #138.

The old example was overwriting previous `param_n` iterations so I fixed that.

`Rosenbrock` in the example could also allow any RNG if we remove the `Default` implementation.

### DIFF
--- a/examples/simulatedannealing.rs
+++ b/examples/simulatedannealing.rs
@@ -14,7 +14,6 @@ use rand::distributions::Uniform;
 use std::default::Default;
 use std::sync::{Arc, Mutex};
 
-#[derive(Clone)]
 struct Rosenbrock {
     /// Parameter a, usually 1.0
     a: f64,
--- a/examples/simulatedannealing.rs
+++ b/examples/simulatedannealing.rs
@@ -79,8 +79,7 @@ impl ArgminOp for Rosenbrock {
             param_n[idx] += val;
 
             // check if bounds are violated. If yes, project onto bound.
-            param_n[idx] = param_n[idx].min(self.upper_bound[idx]);
-            param_n[idx] = param_n[idx].max(self.lower_bound[idx]);
+            param_n[idx] = param_n[idx].clamp(self.lower_bound[idx], self.upper_bound[idx]);
         }
         Ok(param_n)
     }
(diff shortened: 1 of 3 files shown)


TheIronBorn:
oh huh, never worked with a `cargo fmt --check` CI. It wants me to move `use rand::distributions::Uniform;` up to join the other rand stuff I guess

stefan-k:
Thanks for this PR! I'm sorry about the delayed reply. Seems like the CI is failing on the tests. Could you have a look please? Thanks! :)

TheIronBorn:
Oh I didn't update the test code. I tried to run tests locally but couldn't due to an esoteric ndarray linker error. `tests.rs` is the only file which requires it so this time I disabled it locally.

examples/simulatedannealing.rs
```diff
@@ -64,27 +64,23 @@ impl ArgminOp for Rosenbrock {
...
 
             // check if bounds are violated. If yes, project onto bound.
-            if tmp > self.upper_bound[idx] {
-                param_n[idx] = self.upper_bound[idx];
-            } else if tmp < self.lower_bound[idx] {
-                param_n[idx] = self.lower_bound[idx];
-            } else {
-                param_n[idx] = param[idx] + val;
-            }
+            param_n[idx] = param_n[idx].min(self.upper_bound[idx]);
```
TheIronBorn, line 82: Just realized we could use the newish [`f64::clamp`](https://doc.rust-lang.org/std/primitive.f64.html#method.clamp) function instead

stefan-k: I would also prefer `clamp` as it seems to be the most clear way of achieving this. Thanks for mentioning it, I didn't know about this method.

stefan-k:
Thanks for adapting the tests, the CI passes now. It would be great if you could adapt the example to use `clamp` instead of `min` and `max`. Otherwise it is ready to go. Thanks a lot for your work! :)

stefan-k:
Looks good now. Thanks a lot once more for your work and for improving the English in the comments! :)

PR #142, opened 2021-11-12
Add bulk ArgminOp functions

TheIronBorn:
Addresses #6 

I considered taking an `IntoIterator` rather than a slice but rayon can't turn that into a parallel iterator without extra trait requirements.

We could update `ParticleSwarm` to use `bulk_apply` like so, but it requires two extra `Vec` collects
```rust
    p.position = O::Param::min(
        &O::Param::max(&new_position, &self.search_region.0),
        &self.search_region.1,
    );
}

let new_costs = op.bulk_apply(self.particles.iter().map(|p| p.position.clone().collect()))?;

for (p, new_cost) in self.particles.iter_mut().zip(new_costs) {
    if new_cost < p.best_cost {
        p.best_position = p.position.clone();
```
If we do it would be good to add a note to the PS example.


stefan-k:
(Sincere apologies once again for having to wait so long for a response!) 

Thanks for your effort, I really appreciate it! 
Is it possible that one of the closing brackets in your example is misplaced in `self.particles.iter().map(|p| p.position.clone().collect())`? I think it should be `self.particles.iter().map(|p| p.position.clone()).collect()`.

Having to clone all positions is somewhat unfortunate efficiency-wise. An `IntoIterator` solution (as you mentioned) would be nicer. Could you elaborate on what the extra trait requirements would be, please?

TheIronBorn:
rayon doesn't have a generic `IntoParallelIterator` impl we could derive from a generic iterator so we would need to add a `where I: IntoParallelIterator` to the trait something like this:
```rust
fn bulk_apply<I>(&self, params: I) -> Result<Vec<Self::Output>, Error>
where
    I: IntoParallelIterator<Item = Self::Param>,
{
    params.into_par_iter().map(|p| self.apply(&p)).collect()
}
```
but that seems to have some extra trait problems as well

stefan-k:
I see, thanks. I agree that these restrictions aren't what we want. 

In the first post you wrote:

> If we do it would be good to add a note to the PS example.

Could you explain what you mean with this, please? Because at first sight I don't see how this affects the user.

I was playing around with this problem a bit and the only way I could make this work without the need to clone all positions was by using a `Vec<Self::Param>` instead of `&[Self::Param]`:

```rust
use rayon::prelude::*;
use std::fmt::Debug;

type Param = Vec<f64>;

#[derive(Debug)]
struct Particle {
    pub position: Param,
}

pub fn bulk_apply(p: Vec<&Param>) {
    let _pff: Vec<()> = p
        .par_iter()
        .map(|b| {
            println!("{:?}", b);
        })
        .collect();
}

fn main() {
    let particles = vec![
        Particle {
            position: vec![0.0, 0.1],
        },
        Particle {
            position: vec![1.0, 4.3],
        },
    ];
    bulk_apply(particles.iter().map(|p| &p.position).collect());
}
```

I'm not sure if this has any performance or other implications.

TheIronBorn:
For the example I mean that we could implement `bulk_apply` directly or just add a note that it might be worthwhile for expensive `apply` functions

stefan-k:
Ah, I see. Implementing `bulk_apply` sounds like a good idea. How do you feel about two examples `particleswarm` and `particleswarm_rayon`?

TheIronBorn:
That would of course require adding rayon as a dev-dependency.

I found that if we do `fn bulk_apply(&self, params: &[&Self::Param]) ->` we can avoid the cloning as well:
```rust
let positions: Vec<_> = self.particles.iter().map(|p| &p.position).collect();
let new_costs = op.bulk_apply(&positions)?;
```
It also aligns better with `fn apply(&self, &Self::Param)`. I should have considered that.

stefan-k:
> That would of course require adding rayon as a dev-dependency.

I don't mind. As far as I know one can define dev-dependencies per example, which is something that I will look into in the near future.

> I found that if we do `fn bulk_apply(&self, params: &[&Self::Param]) ->` we can avoid the cloning as well

This is great! I think this is the most convenient solution.

(Just a sidenote: I will be mostly offline for multiple days starting tomorrow, therefore I most likely won't be able to respond here for a while.)

TheIronBorn:
I've realized that the ref version (`&[&self::Param]`) makes other solvers harder:
```rust
let new_params = op.bulk_modify(&self.params.iter().collect::<Vec<_>>(), &self.temps)?;
let new_costs = op.bulk_apply(new_params)?;
```

I suppose it's an array-of-structs vs struct-of-arrays question.

Though a more direct implementation of Particle Swarm would avoid extra 'collect's altogether:
```rust
let inner_op = op.op.as_ref().unwrap();
op.cost_func_count += self.particles.len() as u64;

self.particles
    .par_iter_mut()
    .try_for_each::<_, Result<_, Error>>(|p| {
        /* ... snip ... */

        p.cost = inner_op.apply(&p.position)?;

        // snip

        Ok(())
    })?;
```
though that changes the scope of this significantly.

I am willing to write a PR for Particle Swarm but I don't know how able other potential solvers are to allow this sort of parallelism.

stefan-k:
> I've realized that the ref version (`&[&self::Param]`) makes other solvers harder:
> 
> ```rust
> let new_params = op.bulk_modify(&self.params.iter().collect::<Vec<_>>(), &self.temps)?;
> let new_costs = op.bulk_apply(new_params)?;
> ```
> 
> I suppose it's an array-of-structs vs struct-of-arrays question.

Which type is `self.params` in your example? 

Btw, a bit offtopic, but what if `bulk_modify` would take the parameters as `&[&mut self::Param]` and return `Result<(), Error>`? Similarily `modify` would also take `&mut self::Param`. I'm not sure if I've considered this when implementing the `ArgminOp` trait, but from my current point of view this seems to be the more efficient. 


> Though a more direct implementation of Particle Swarm would avoid extra 'collect's altogether:
> 
> ```rust
> let inner_op = op.op.as_ref().unwrap();
> op.cost_func_count += self.particles.len() as u64;
> 
> self.particles
>     .par_iter_mut()
>     .try_for_each::<_, Result<_, Error>>(|p| {
>         /* ... snip ... */
> 
>         p.cost = inner_op.apply(&p.position)?;
> 
>         // snip
> 
>         Ok(())
>     })?;
> ```
> 
> though that changes the scope of this significantly.
> 
> I am willing to write a PR for Particle Swarm but I don't know how able other potential solvers are to allow this sort of parallelism.

If I understand this correctly, this would require the rayon dependency. I would like to avoid this, even though I think that what you suggested is much more elegant.

TheIronBorn:
> Which type is self.params in your example?

It's `Vec<Self::Param>`.

Many solvers which need `modify` need continued access to the previous/current parameter but we can get that from `IterState` so I think it's doable.

stefan-k:
> It's `Vec<Self::Param>`.

I see, so `&self.params.iter().collect::<Vec<_>>()` is to get from `Vec<Self::Param>` to `&Vec<&Self::Param>` which can be passed to `modify` which expects `&[&Self::Param]`. It's a bit clunky but I guess one can maybe adapt the affected solvers such that this is avoided. And even if it can't be avoided I guess (read: hope) the compiler will still be able to produce efficient code. 

> Many solvers which need `modify` need continued access to the previous/current parameter but we can get that from `IterState` so I think it's doable.

Is this in response to the `&mut` suggestion? I think I wouldn't modify the parameters stored in `IterState` directly because the idea is that `IterState` is solely managed by the `Executor`.

Btw, thank you for the huge effort you have put into this already. I really appreciate it!

stefan-k:
You may have to merge the current master branch into your branch before pushing again. The CI broke in the meantime and the fix is in master.

stefan-k:
I was wondering if this:

> ```rust
> let new_params = op.bulk_modify(&self.params.iter().collect::<Vec<_>>(), &self.temps)?;
> let new_costs = op.bulk_apply(new_params)?;
> ```

could be rewritten as this:

> ```rust
> let new_params = op.bulk_modify(&self.params.iter().map(AsRef::as_ref).collect(), &self.temps)?;
> let new_costs = op.bulk_apply(new_params)?;
> ```

I think this would indicate better what the intention is. 

However, looking at it more closely, I realized that I'm a bit unhappy about how `modify` is designed in general. It should either take a mutable reference to the parameter vector such that it can be modified directly OR the parameter vector should be moved, modified, and returned by the method again. Using non-mut references forces one to create a new parameter vector, which may be inefficient. Do you have any opinions on this? At this point I feel like moving might be the preferable option. In the particular case of PSO this may be suboptimal, but my feeling is that PSO might require a redesign anyways.

stefan-k:
> Though a more direct implementation of Particle Swarm would avoid extra 'collect's altogether:
> 
> ```rust
> let inner_op = op.op.as_ref().unwrap();
> op.cost_func_count += self.particles.len() as u64;
> 
> self.particles
>     .par_iter_mut()
>     .try_for_each::<_, Result<_, Error>>(|p| {
>         /* ... snip ... */
> 
>         p.cost = inner_op.apply(&p.position)?;
> 
>         // snip
> 
>         Ok(())
>     })?;
> ```
> 
> though that changes the scope of this significantly.
> 
> I am willing to write a PR for Particle Swarm but I don't know how able other potential solvers are to allow this sort of parallelism.

Now that I've spent more time with PSO in #216 (PSO wasn't initially written by me), I came to the conclusion that I kind of like your suggestion quoted above the most. But it would be really nice if users could somehow turn this feature on and off as they wish (and hence avoid the rayon dependency). I just don't know how.

TheIronBorn:
Wouldn't a feature flag work? 
```rust
#[cfg(feature = "rayon")] // or some custom feature
{ /* do proposed method */ }

#[cfg(not(feature = "rayon"))]
{ /* do regular */ }

stefan-k:
Sure. My initial thought though was that users should be able to choose the sequential version even if they have rayon activated, and I wasn't sure how to do that properly while avoiding unnecessary trait bounds. Anyway, I thought about it in more detail today and came to the conclusion that 

a) anything but "rayon on == parallelization" and "rayon off == sequential" is probably unnecessarily confusing and
b) implementing the bulk_ methods is probably still the best approach as it is also the most future-proof. 

To avoid further mood swings of myself between all available options I decided to write some code to see what would feel the best. After I figured out how to handle parallel iterators with generics it boiled down to one of your suggestions. That of course leads to `Send` and `Sync` bounds all over the place which causes unnecessary limits for users who want to run their code sequentially. I solved this in the same way as I solved it with the `Serialize` and `DeserializeOwned` traits: By implementing alias traits. 

I was essentially just playing around, but since the proof of concept works (#219), all thats missing is a bit of copy-paste and writing docs. 

Apologies that I "stole" your PR, but given how far #219 is and how much of argmin has changed since work on this PR started, I figured that pushing my code is much less work than rebasing this PR to the current state of argmin.

Thanks a lot for your work in this PR and for the valuable discussions as well as your patience when I didn't respond within reasonable time. I really appreciate it!

TheIronBorn:
No worries. Glad to help?

Issue #151, opened 2021-11-28
`ObserverMode::NewBest` sometimes logs everything


Glitchy-Tozier:
Originated in #150.

> I assume this is likely due to differences in cost, which are beyond the digits that are printed to the screen. If your changes in the `modify` function are scaled with your small temperature, your movements will be small and therefore also the differences in cost are likely to be very small.

They are not.

In case it helps you in any way, here's the `modify()`-function (the important thing to note is that `1` in switch_n_keys`. It always is the same):

```rs
/// This function is called by the annealing function
fn modify(&self, param: &Self::Param, _temp: f64) -> Result<Self::Param, Error> {
    // in the following, the 1 will be replaced by something `temp`-depending
    Ok(self.layout_generator.switch_n_keys(&param, 1))
}
```

### Here's a new weird thing, that might help us here and/or in #150:
I have modified the `apply()`-function so that the cost also gets printed:
```rs
fn apply(&self, param: &Self::Param) -> Result<Self::Output, Error> {
    let layout = self.layout_generator.generate_layout(&param);
    let evaluation_result = self.evaluator.evaluate_layout(&layout);
    println!(
        "\nprintln from inside `apply()`: Cost: {}",
        evaluation_result.total_cost()
    );
    Ok(evaluation_result.total_cost())
}
```

This led to some interesting results:

If everything works fine (except for #150, of course), this now is the output:
```rs
Nov 28 21:17:18.569 INFO Simulated Annealing, max_iters: 10000, initial_temperature: 0.01, stall_iter_accepted_limit: 1000, stall_iter_best_limit: 18446744073709551615, reanneal_fixed: 18446744073709551615, reanneal_accepted: 18446744073709551615, reanneal_best: 18446744073709551615

println from inside `apply()`: Cost: 302.38946704551745
Nov 28 21:17:18.730 INFO iter: 0, cost: 302.38946704551745, best_cost: 302.38946704551745, cost_func_count: 1, grad_func_count: 0, jacobian_func_count: 0, hessian_func_count: 0, modify_func_count: 1, t: 0.014426950408889635, new_be: true, acc: true, st_i_be: 0, st_i_ac: 0, ra_i_fi: 1, ra_i_be: 0, ra_i_ac: 0, ra_fi: false, ra_be: false, ra_ac: false, time: 0.16084857200000002

println from inside `apply()`: Cost: 354.9768532415309

println from inside `apply()`: Cost: 410.09772736186335

[...]

PR #152, opened 2021-12-01
Correct check for new best parameter

stefan-k:
addresses #151


stefan-k:
CI is suddenly failing at compiling `openblas-src`. I will check back tomorrow, hopefully it fixed itself by then.

Issue #150, opened 2021-11-28
Simulated Annealing: First mutation always gets accepted?


Glitchy-Tozier:
Hello, I have encountered a scenario which I don't quite understand.

I tried to set the initial temperature (in Simulated annealing) to the smallest possible positive number. (See first line of code-block) According to my current understanding, This should basically result in a greedy hill-climbing algorithm, where only better parameters are accepted.

This isn't fully the case though. The first modification/mutation always seems to get accepted, even though it is way worse.

Code containing the `init_temp`:
```rs
let init_temp = f64::MIN_POSITIVE;

let problem = AnnealingStruct {
    evaluator: Arc::new(evaluator.clone()),
    layout_generator: &pm,
};

// Create new SA solver with some parameters (see docs for details)
// This essentially just prepares the SA solver. It is not run yet, nor does it know anything about the problem it is about to solve.
let solver = SimulatedAnnealing::new(init_temp) // 200.0)
    .unwrap()
    // Optional: Define temperature function (defaults to `SATempFunc::TemperatureFast`)
    .temp_func(SATempFunc::Boltzmann)
    /////////////////////////
    // Stopping criteria   //
    /////////////////////////
    // Optional: stop if there was no accepted solution after 1000 iterations
    .stall_accepted(params.stall_accepted);

// Create and run the executor, which will apply the solver to the problem, given a starting point (`init_param`)
let res = Executor::new(problem, solver, init_layout)
    // Optional: Attach a observer
    .add_observer(ArgminSlogLogger::term(), ObserverMode::Always) //Every(100))
    // Optional: Set maximum number of iterations (defaults to `std::u64::MAX`)
    .max_iters(params.max_iters)
    // Optional: Set target cost function value (defaults to `std::f64::NEG_INFINITY`)
    .run()
    .unwrap();
```

Log:
```rs
Nov 28 16:53:58.373 INFO Simulated Annealing, max_iters: 10000, initial_temperature: 0.000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000022250738585072014, stall_iter_accepted_limit: 1000, stall_iter_best_limit: 18446744073709551615, reanneal_fixed: 18446744073709551615, reanneal_accepted: 18446744073709551615, reanneal_best: 18446744073709551615

[...]

PR #153, opened 2021-12-01
[fix] simulated anneal always accepts first iter

dariogoetz:
This addresses #150 

It appears to me that the initial `ArgminIterData` returned by the simulated anneal solver's `init` method does not contain values for `param` and `cost`. Hence the `executor`'s state is not being updated upon calling the `solver`'s `init` method. This leads to the state having infinity as starting cost, which makes any first solution preferable.

I have added param and corresponding initial cost to the solver's init method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## argmin unreleased (xx xxxxxx xxxx)
+
+- Fixed simulated annealing always accepting the first iteration (#153, @dariogoetz)
+
 ## argmin v0.4.7 (14 August 2021)
 
 - Moved to Github actions (#130, @stefan-k)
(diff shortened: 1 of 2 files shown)


stefan-k:
Thank you for this PR! This is actually what I also suspected, apologies that I wasn't able to address this  myself.

For some reason the CI is broken since yesterday. I'm having the same problem in #152. I will try to have a look hopefully this evening. My suspicion is that something in the images changed.

dariogoetz:
Don't worry, it was not my intention to apply any pressure to you. I just wanted to offer my help with the PR.

stefan-k:
I did not interpret this as pressure :) Any help is highly appreciated.

You're welcome to update the changelog if you want, otherwise I will update it once I prepare the next release. 

Seems like I was able to fix the CI problems. If you merge the current master into your branch then the compilation of  `openblas-src` should work again and the clippy errors should be gone.

dariogoetz:
Thanks for fixing the CI. I have updated the changelog, adding a line about the fix.

stefan-k:
Thanks a lot once more for this PR!

Issue #154, opened 2021-12-03
Issues with backtracking line search


w1th0utnam3:
I noticed some points while playing around with LBFGS in combination with backtracking line search and would like to hear your thoughts before submitting a PR.

- I think it would make sense if the optimizer (e.g. LBFGS in this case) accumulated the cost function, gradient etc. call counts into its own call counts because these are the numbers that are reported to an observer. Currently the reported numbers are misleading because the functions are actually called much more often depending on the line search.
- Currently the backtracking line search evaluates its condition (in [`terminate`](https://github.com/argmin-rs/argmin/blob/f5ffe75979f5d14574e666837bd1220c2e28e727/src/solver/linesearch/backtracking.rs#L163)) at iteration zero (i.e. directly after the call to `init` and before the first call to `next_iter`) without actually applying the search direction to the current state. This doesn't make any sense to me.
- The `alpha` values in backtracking line search are actually "shifted" by one application of `rho` when passed into the condition. For example in the first iteration (`iter == 1`) it actually checks the condition with a cost and gradient evaluated at `x + alpha * search_direction` (where `alpha == initial_alpha`) but at the same time supplies the value [`alpha = initial_alpha * rho`](https://github.com/argmin-rs/argmin/blob/f5ffe75979f5d14574e666837bd1220c2e28e727/src/solver/linesearch/backtracking.rs#L150) to the condition. This appears to be inconsistent.

I'm not sure if these issues apply to other line searches as well. If you agree that these points should be addressed I can submit a PR.

stefan-k:
> * I think it would make sense if the optimizer (e.g. LBFGS in this case) accumulated the cost function, gradient etc. call counts into its own call counts because these are the numbers that are reported to an observer. Currently the reported numbers are misleading because the functions are actually called much more often depending on the line search.

This is actually what `op.consume_op(line_op);` in [line 162 of LBFGS is supposed to do](https://github.com/argmin-rs/argmin/blob/f5ffe75979f5d14574e666837bd1220c2e28e727/src/solver/quasinewton/lbfgs.rs#L162). I'm surprised that this doesn't work. 

[...]

PR #155, opened 2021-12-03
Fix inconsistency between state and alpha value in Backtracking linesearch

w1th0utnam3:
This fixes an inconsistency between the state dependent values and the alpha value (which was one iteration ahead of the state) supplied to the line search condition in `Solver::terminate`.

I also removed the trait bounds `ArgminSub` and `ArgminDot` on the `Solver` impl as they were not used. I'm not sure if this was intentional.

See https://github.com/argmin-rs/argmin/issues/154.

### DIFF
--- a/src/solver/linesearch/backtracking.rs
+++ b/src/solver/linesearch/backtracking.rs
@@ -93,15 +93,37 @@ where
     }
 }
 
+impl<P, L, F: ArgminFloat> BacktrackingLineSearch<P, L, F>
+where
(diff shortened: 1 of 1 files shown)


stefan-k (approved):

stefan-k:
Thank you for submitting this PR! 

> I also removed the trait bounds `ArgminSub` and `ArgminDot` on the `Solver` impl as they were not used. I'm not sure if this was intentional.

Good catch, this certainly is not intentional. :)

PR #158, opened 2021-12-07
Update LineSearchCondition trait to use references

w1th0utnam3:
Currently the `LineSearchCondition` takes its vector arguments by value. Consequently, the line search always has to explicitly clone the gradient / search direction vectors. For larger systems this might be expensive, especially if many line search iterations are needed. I don't see a reason why an explicit clone should be enforced by the interface. Therefor, I would propose to change the trait to accept references.

### DIFF
--- a/src/solver/linesearch/backtracking.rs
+++ b/src/solver/linesearch/backtracking.rs
@@ -11,6 +11,7 @@ use crate::prelude::*;
 use crate::solver::linesearch::condition::*;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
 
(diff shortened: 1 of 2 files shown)


w1th0utnam3:
Oops, forgot an `unwrap()` and somehow my `cargo check` didn't catch this without a `cargo clean` first.

stefan-k:
Thanks a lot for this contribution! This was actually on my TODO list, glad to see that this can now be ticked off :) 

Apologies for the failing CI. Since a couple of days it fails randomly on building `openblas-src`. I thought I had fixed it, but obviously didn't. I'll restart the pipeline tomorrow morning in the hope that it will work then.

stefan-k (approved):

PR #184, opened 2022-02-05
Fix #2

### LINKED ISSUE #2: Bug in HagerZhangLineSearch

wfsteiner:
Avoid termination before linesearch starts. Fix in comments


stefan-k:
I don't know if you've seen my reply to your message in the Gitter channel. I just gave it a try and it indeed seems to fix the problem.
Thanks a lot for this contribution! :)

I will try to merge this into a new branch starting off at the `v0.5.0` tag because I intend to release this as a patch release. This is just a heads-up in case something strange happens in this PR (I have never done this before).

Issue #71, opened 2020-09-01
Checkpointing should be configurable by user


stefan-k:
Currently checkpoints can only be [written to disk](https://github.com/argmin-rs/argmin/blob/68c282f54e19fb6b22dc40cfaf78ce861b78842e/src/core/serialization.rs#L64). It would be better if users could write their own checkpointing mechanisms. In order to do so, a checkpointing trait should be defined. This could be implemented similar to [observers](https://github.com/argmin-rs/argmin/blob/master/src/core/observers/mod.rs).

PR #198, opened 2022-02-18
Large refactoring of operators, state, observers, checkpointing, documentation, ...

stefan-k:
This PR is still experimental. It allows for using operators other than `ArgminOp` (for instance `LinearProgram`). It also adds the possibility to build `Solver`-tailored `State`s passed from one iteration to the next.

This may lead to `ArgminOp` being split into multiple traits, one for each trait method. This would make some things more consistent and easier for users. For instance, the `modify` function could become part of an `Anneal` trait in the SA solver and part of a `Move` trait in the PSO solver. 
This would certainly affect #142 .

Feedback is welcome!

EDIT: The scope of this has expanded a bit. For a detailed explanation of what happened here please see the later posts.


stefan-k:
The scope of this PR has expanded "a bit". Rough overview of what happened:

* Many aspects of the documentation were improved and many doc tests were added.
* The structuring of the library in modules was changed substantially.
* The `visualizer` feature was removed (the code was somewhat limited and not maintained). Instead the observer code was moved to an example to showcase how observers can be implemented.
* The trait `ArgminOp` was removed and replaced by the traits `CostFunction`, `Operator`, `Gradient`, `Jacobian`, `Hessian` and `Anneal`. Solvers now define exactly which implementations are needed. Solvers can also require special traits only relevant to the said solver (such as `Anneal` which is only relevant for `SimulatedAnnealing`).
* It is now possible to use different internal states when implementing solvers. In addition to `IterState` there is now also `LinearProgramState` (so far unused). Common needed methods are defined in the `State` trait. Initialization of a solver (such as providing an initial parameter vector) is now done via explicitly calling methods of the state. This can be done with `Executor`s `configure(...)` method which accepts a closure. Please see the examples or the documentation for how to use this. 
* Checkpointing is now provinding a trait such that users can implement their own checkpointing algorithms. Checkpointing and the configuration of an Executor using checkpointing was greatly simplified. Please see the example or the documentation for details. (fixes #71 )
* `CheckpointMode` was renamed to `CheckpointingFrequency`
* When starting from a checkpoint, the solver is not initialized anymore (fixes #199). A regression test was written.
* `OpWrapper` was renamed to `Problem` and keeps track of function evaluations via a `HashMap` (and can therefore also work with "solver defined" traits which the problem needs to implement.
* `ArgminIterData` was removed, instead the current `state` and `kv` are returned.
* Solvers can now directly modify the state
* Macros are now all in a single file.
* `ArgminKV` was renamed to `KV`
* `KV` now stores `Rc<dyn Display>` values instead of strings. This has the advantage that formatting the values to strings is only done when actually using them in a logger. In the past this was always done, even if there was not logger attached, leading to a huge performance drop.
* `ArgminLineSearch` was renamed to `LineSearch` and moved to line search module
* `ArgminNLCGUpdate` was renamed to `NLCGUpdate` and moved to CG module
* `ArgminTrustRegion` was renamed to `TrustRegion` and moved to trustregion module
* The `NoOperator`s were renamed to `TestProblem`.
* A `TestSolver` was introduced for testing.
* The methods of `Observe` get mutable access to `self`
* `Observer` was renamed to `Observers` (it is a container of observers after all)
* `Observers` is now tested
* `ArgminSlogLogger` was renamed to `SlogLogger`
* `ArgminResult` was renamed to `OptimizationResult` and contains the final `state`, `problem` and `solver`. (And when printed it shows which solver was used).

I probably missed something. I will merge this PR now since it is already way to big. `IterState` will be cleaned up and documented in a different PR. The solvers will also be documented later on.

PR #225, opened 2022-06-11
CMA-ES

VolodymyrOrlov:
Implements [CMA-ES](https://en.wikipedia.org/wiki/CMA-ES) algorithm, as described in [this paper](https://arxiv.org/pdf/1604.00772.pdf). Along with CMA-ES, this PR extends `argmin-math` with additional functions:
- argsort: https://numpy.org/doc/stable/reference/generated/numpy.argsort.html
- eigSym: https://www.rdocumentation.org/packages/MGMM/versions/0.3.1/topics/eigSym
- np.random.standard_normal: https://numpy.org/doc/stable/reference/random/generated/numpy.random.standard_normal.html
- np.outer: https://numpy.org/doc/stable/reference/generated/numpy.outer.html
- np.take: https://numpy.org/doc/stable/reference/generated/numpy.take.html


stefan-k:
This is huge! Thanks a lot, not just for CMA-ES, but also for the massive extension of argmin-math, which will certainly come in handy in the future. 
Since this is quite a large PR, I'll have to take some time to review it. I hope to be able to review it in the course of the next week.

VolodymyrOrlov:
Sounds good, @stefan-k ! Thanks for taking a look at the PR!

VolodymyrOrlov:
Thank you for giving it a try, @Armavica! What backend are you using, vec/ndarray or nalgebra?

Armavica:
I just minimally modified the example in [`cmaes.rs`](https://github.com/VolodymyrOrlov/argmin/blob/main/argmin/examples/cmaes.rs) and ran it with `cargo run --release --example cmaes`, so I think it is `vec` unless I am wrong?

Armavica:
It looks like there is a similar problem with the `nalgebra` backend, observed if you change the maximum number of iterations from `100` to `10000` in `test_nalgebra_solver`?

argmin-math/src/lib.rs
```diff
@@ -275,3 +279,102 @@ pub trait ArgminMinMax {
     /// Select piecewise maximum
     fn max(x: &Self, y: &Self) -> Self;
 }
+
+/// Defines associated types
+pub trait ArgminTransition {
```
stefan-k, line 284: It is not entirely clear to me what this does and what it is needed for. Could you elaborate a bit, please?

argmin-math/src/lib.rs
```diff
@@ -275,3 +279,102 @@ pub trait ArgminMinMax {
...
+
+/// Take elements from an array along an axis.
+pub trait ArgminTake<I> {
+    /// Take elements from indices along a given axis.
+    /// Returns as new array of the same size as indices
+    fn take(&self, indices: &[I], axis: u8) -> Self;
+}
+
+/// Defines iterator
+pub trait ArgminIter<T> {
```
stefan-k, line 324: From the implementation it seems to me that instead of `ArgminIter` (and `ArgminMutIter`), one could maybe use `Iterator` directly. However, I do feel that this way it is more consistent with `ArgminAxisIter`, which is more complicated and hence makes sense to abstract over for different backends. I guess I'm just thinking out lout because I'm not sure what I want to say with this comment. I don't think that this should be changed.

argmin-math/src/vec/shur.rs
```diff
@@ -0,0 +1,382 @@
...
+// copied, modified, or distributed except according to those terms.
+
+#![allow(non_snake_case)]
+use crate::vec::HessenbergTransformer;
+
+const SHUR_MAX_ITERATIONS: usize = 100;
+
+/// Transforms a matrix into a Shur form.
+/// This transformation is an intermediate step to Eigendecomposition
+pub(crate) struct ShurTransformer<T> {
```
stefan-k, line 15: Isn't it "Schur" instead of "Shur"? Maybe I'm thinking of something different.

stefan-k (changes requested):
I only have a few minor comments. Unfortunately I'm not a user of `nalgebra` therefore I'll have to trust you on that ;)

argmin-math/src/vec/shur.rs
VolodymyrOrlov: I think you are correct, it should be `Schur`

argmin-math/src/lib.rs
VolodymyrOrlov: I need to some mechanism that will return me a type of 2-D array in exchange for 1-D array. Take a look at [this line of code](https://github.com/argmin-rs/argmin/blob/e4bb91a84fd3f2d5173524bf8b3916f06734a046/argmin/src/solver/cma_es/mod.rs#L135). To create an instance of 2-D identity matrix I need to know a type of the matrix. The problem is all my parameters are either scalars or 1-D vectors. This class helps me to transition from 1-D to 2-D array.

argmin-math/src/lib.rs
VolodymyrOrlov: I don't think it will be safe to assume that every single type of 1-D array will implement `Iterator` trait. `vec` doesn't.

VolodymyrOrlov:
> I only have a few minor comments. Unfortunately I'm not a user of `nalgebra` therefore I'll have to trust you on that ;)
> 
> After addressing these comments and the problems @Armavica pointed out I think this is ready to go!

Thank you for reviewing my PR! I've found a couple of numerical issues that lead to problems pointed out by @Armavica! I hope to have some changes for you guys some time by the mid next week.

argmin-math/src/lib.rs
stefan-k: Thanks for the explanation, makes sense!

argmin-math/src/lib.rs
stefan-k: Agree!

VolodymyrOrlov:
I've fixed issues discovered by @Armavica . The vector-based implementation is still inferior to `ndarray` and `nalgebra`, but at least the method converges for multidimensional Rosenbrock and doesn't panic for when number of iterations exceeds 100

stefan-k (changes requested):
> I've fixed issues discovered by @Armavica . The vector-based implementation is still inferior to ndarray and nalgebra, but at least the method converges for multidimensional Rosenbrock and doesn't panic for when number of iterations exceeds 100

Thanks a lot once again! :) Just out of curiosity, in which way is the vec-based implementation inferior to the others? 

I only have a very few minor comments. One maybe larger one is the use of `bulk_cost` instead of `cost` which enables parallel computation of the cost function for the entire population. Other than that I think this is good to go. Before merging, could you squash your commits a bit please? It doesn't need to be a single commit but grouping them in a couple of commits would be nice.

argmin-math/src/vec/schur.rs
```diff
@@ -0,0 +1,382 @@
+// Copyright 2018-2022 argmin developers
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+#![allow(non_snake_case)]
```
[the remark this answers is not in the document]
VolodymyrOrlov: yes, Clippy won't let me have non camel-case names otherwise.

VolodymyrOrlov:
> > I've fixed issues discovered by @Armavica . The vector-based implementation is still inferior to ndarray and nalgebra, but at least the method converges for multidimensional Rosenbrock and doesn't panic for when number of iterations exceeds 100
> 
> Thanks a lot once again! :) Just out of curiosity, in which way is the vec-based implementation inferior to the others?
> 
> I only have a very few minor comments. One maybe larger one is the use of `bulk_cost` instead of `cost` which enables parallel computation of the cost function for the entire population. Other than that I think this is good to go. Before merging, could you squash your commits a bit please? It doesn't need to be a single commit but grouping them in a couple of commits would be nice.

The vector and nalgebra backend requires more iterations to converge because Eigenvector decomposition implementation is not as good a the one that comes with LAPACK

You will notice that number of modified files went up with my latest changes. This is due to commits squashing.

stefan-k:
> The vector and nalgebra backend requires more iterations to converge because Eigenvector decomposition implementation is not as good a the one that comes with LAPACK

Thanks for the explanation!
 
> You will notice that number of modified files went up with my latest changes. This is due to commits squashing.

I think something went wrong during squashing. It seems that merging main into your branch (rather than rebasing) caused an unfortunate starting point for squashing. Most commits were squashed into one of mine, which led to them being assigned to my user. I tried to fix this, so now I'm kindly asking you to review the resulting changes once the CI is successful ;) 
I saved the previous commits just in case I messed up.

VolodymyrOrlov:
> > The vector and nalgebra backend requires more iterations to converge because Eigenvector decomposition implementation is not as good a the one that comes with LAPACK
> 
> Thanks for the explanation!
> 
> > You will notice that number of modified files went up with my latest changes. This is due to commits squashing.
> 
> I think something went wrong during squashing. It seems that merging main into your branch (rather than rebasing) caused an unfortunate starting point for squashing. Most commits were squashed into one of mine, which led to them being assigned to my user. I tried to fix this, so now I'm kindly asking you to review the resulting changes once the CI is successful ;) I saved the previous commits just in case I messed up.

Thank you for your help with squashing commits, @stefan-k ! The changes looks good, feel free to merge this PR when you deem it ready.

Armavica:
Hi, I pulled the changes in the branch `VolodymirOrlov/main` and modified the `examples/cmaes.rs` file to allow for 1000 iterations. The optimization still panics. Am I doing something wrong?

stefan-k:
> Hi, I pulled the changes from the branch `VolodymirOrlov/main` and modified the `examples/cmaes.rs` file to allow for 1000 iterations. The optimization still panics. Am I doing something wrong?

Interestingly, I have no issues with 1000 iterations, but I do see a panic for 2000 iterations: 

```
thread 'main' panicked at 'Exceeded maximum number of iterations when calculating QL transformation', /home/stefan/coding/argmin/argmin/argmin-math/src/vec/eigen.rs:422:1
```

Armavica:
Ah yes indeed sorry, I was apparently doing something wrong: it also works most of the time for me now. I also get this panic sometimes, non-reproducibly. From a few rapid tests this implementation seems to work even even for high-dimensional functions. Congrats for fixing the issue!

The author of CMA-ES has a page with a few different examples to help test and benchmark CMA-ES implementations: https://cma-es.github.io/cmaes_sourcecode_page.html#testing I think it could be nice ideally to see how this implementation performs on these tests, but maybe this can come later…

stefan-k:
From a quick look I've noticed that there are some direct comparisons of floats, for instance:

```rust
if (e[i + 1] == 0.0) {
    eigenvalues[i + 1] -= u;
    e[m] = 0.0;
    break;
}
```
and

```rust
if t == 0. && i >= j {
    continue;
}
```
Could this cause issues because it is unlikely that those will be exactly 0?

stefan-k:
Hi @VolodymyrOrlov , this PR is really far ahead and I'd really like to include it in the next release. There only seem to be one remaining issue regarding the Eigenvalue decomposition for the vector and nalgebra backends and the PR would have to be rebased onto the current main. I'm unsure how to proceed. I wouldn't mind dropping support for the `Vec` backend as it is rather limited and inefficient anyways, but it would be sad if all the tremendous effort you put into this would be for nothing.
Support for the `nalgebra` backend however would be pretty great. What do you suggest?

VolodymyrOrlov:
> Hi @VolodymyrOrlov , this PR is really far ahead and I'd really like to include it in the next release. There only seem to be one remaining issue regarding the Eigenvalue decomposition for the vector and nalgebra backends and the PR would have to be rebased onto the current main. I'm unsure how to proceed. I wouldn't mind dropping support for the `Vec` backend as it is rather limited and inefficient anyways, but it would be sad if all the tremendous effort you put into this would be for nothing. Support for the `nalgebra` backend however would be pretty great. What do you suggest?

Hi @stefan-k! Thanks for continued support of this PR and my apologies for not acting sooner on suggested changes. I switched companies and now work for a small startup. This consumes most of my time and I didn't had a chance to do any open-source work lately. Give me some time tomorrow and early next week - I'd like to try to resolve remaining issues and update PR to prepare it for the merge.

stefan-k:
Rereading my previous message again I realized that it may have come across as pushy. I apologize if it made that impression, this was certainly not my intention. 
Thanks for continuing to work on this! Take your time, I know how difficult it can be to manage both work and open source contributions. Congratulations on the new job! :)

Boscop:
@VolodymyrOrlov Thanks for your great work, I'm looking forward to this being merged :)

PR #231, opened 2022-07-15
Adding ndarray implementations for ArgminRandom and ArgminMinMax

hypotrochoid:
Adds missing trait implementations for ndarray inputs.

### DIFF
--- a/argmin-math/src/ndarray_m/mod.rs
+++ b/argmin-math/src/ndarray_m/mod.rs
@@ -20,6 +20,7 @@ mod l1norm;
 mod l2norm;
 mod minmax;
 mod mul;
+mod random;
 mod scaledadd;
 mod scaledsub;
 mod signum;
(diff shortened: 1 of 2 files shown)


stefan-k:
Thanks for the contribution! Could you have a look at the clippy lints and run rustfmt, please? Other than that it looks good to me :)

stefan-k:
On other thing I noticed: Could you add tests, please? I know that the other backends are also missing tests, but this could be a good place to start testing this ;) One thing I could imagine in case of `ArgminRandom` is checking that the values are within the provided bounds.

stefan-k:
The remaining clippy errors are due to an update of clippy. You can simply add `Eq` as suggested in the logs.

stefan-k:
I rebased this commit and removed the ndarray ArgminMinMax implementation in favor of the one which was added to main since this PR was opened. I also implemented the trait directly for all basic types instead of using traits because this causes fewer problems down the line in my experience. 

I am merging this PR despite the missing tests which will be added later. 

Thanks again for this contribution! :)

Issue #222, opened 2022-05-14
Implement Powell's method


nikmas-dev:
It would be great to have the Powell algorithm like in scipy optimize: https://docs.scipy.org/doc/scipy/reference/optimize.minimize-powell.html#optimize-minimize-powell

stefan-k:
Thanks for opening this issue! This would be an ideal initial project for someone interested in contributing. I would of course provide mentoring/support if needed.

Trombach:
Hi @stefan-k, came across this library and would love to contribute. I'll have a look at this when I have a bit of time to spare.

stefan-k:
Great! :) Feel free to open a PR when you're ready or when there is something to discuss.

PR #234, opened 2022-07-22
[DRAFT] Powell's algorithm

Trombach:
Hi @stefan-k,

I'm opening this PR to seek some guidance and feedback on my first draft to implement Powell's algorithm (#222).

Currently, this code is completely untested as I am only trying to gauge whether my attempt on implementing this solver is going in the right direction.

Here are some questions I have:
1. Is my approach sensible at all? What can be improved?
2. AFAIK, Powell's method can also be implemented using Golden-section search or Brent's method. Is there a way to do this generically in argmin or will there have to be different versions of solvers for each variant?
3. I would appreciate some suggestions on a good test function to write some tests for this solver.

Thanks!

### DIFF
--- a/argmin/src/solver/powell/mod.rs
+++ b/argmin/src/solver/powell/mod.rs
@@ -1,0 +1,97 @@
+use crate::core::{
+    ArgminFloat, CostFunction, DeserializeOwnedAlias, Executor, IterState, LineSearch,
+    OptimizationResult, SerializeAlias, Solver, State,
+};
+use argmin_math::{ArgminAdd, ArgminDot, ArgminSub, ArgminZeroLike};
+#[cfg(feature = "serde1")]
+use serde::{Deserialize, Serialize};
+use std::mem;
+
+#[derive(Clone)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+pub struct PowellLineSearch<P, L> {
+    search_vectors: Vec<P>,
+    linesearch: L,
+}
+
+impl<P, L> PowellLineSearch<P, L> {
+    pub fn new(initial_search_vectors: Vec<P>, linesearch: L) -> Self {
+        PowellLineSearch {
+            search_vectors: initial_search_vectors,
+            linesearch,
+        }
+    }
+}
+
+impl<O, P, F, L> Solver<O, IterState<P, (), (), (), F>> for PowellLineSearch<P, L>
+where
+    O: CostFunction<Param = P, Output = F>,
+    P: Clone
+        + SerializeAlias
+        + DeserializeOwnedAlias
+        + ArgminAdd<P, P>
+        + ArgminZeroLike
+        + ArgminSub<P, P>
+        + ArgminDot<P, F>,
+    F: ArgminFloat,
+    L: Clone + LineSearch<P, F> + Solver<O, IterState<P, (), (), (), F>>,
+{
+    const NAME: &'static str = "Powell-LS";
+
+    fn next_iter(
+        &mut self,
+        problem: &mut crate::core::Problem<O>,
+        mut state: IterState<P, (), (), (), F>,
+    ) -> Result<(IterState<P, (), (), (), F>, Option<crate::core::KV>), anyhow::Error> {
+        let param = state
+            .take_param()
+            .ok_or_else(argmin_error_closure!(NotInitialized, "not initialized"))?; // TODO add Error message
+
+        // new displacement vector created from displacement vectors of line searches
+        let new_displacement = param.zero_like();
+        let mut best_direction: (usize, F) = (0, float!(0.0));
+
+        // init line search
+        let (ls_state, _) = self.linesearch.init(problem, state.clone())?;
+
+        // Loop over all search vectors and perform line optimization along each search direction
+        for (i, search_vector) in self.search_vectors.iter().enumerate() {
+            self.linesearch.search_direction(search_vector.clone());
+
+            let line_cost = ls_state.get_cost();
+
+            // Run solver
+            let OptimizationResult {
+                problem: _sub_problem,
+                state: sub_state,
+                ..
+            } = Executor::new(problem.take_problem().unwrap(), self.linesearch.clone())
+                .configure(|state| state.param(param.clone()).cost(line_cost))
+                .ctrlc(false)
+                .run()?;
+
+            // update new displacement vector
+            let displacement = &sub_state.get_best_param().unwrap().sub(&param);
+            let displacement_magnitude = displacement.dot(&displacement).sqrt();
+            new_displacement.add(displacement);
+
+            //store index of lowest cost displacement vector
+            if best_direction.1 < displacement_magnitude {
+                best_direction.0 = i;
+                best_direction.1 = displacement_magnitude;
+            }
+        }
+
+        // replace best performing search direction with new search direction
+        let _ = mem::replace(
+            &mut self.search_vectors[best_direction.0],
+            new_displacement.clone(),
+        );
+
+        // set new parameters
+        let param = param.add(&new_displacement);
+        let cost = problem.cost(&param);
+
+        Ok((state.param(param).cost(cost?), None))
+    }
+}
--- a/argmin/src/solver/mod.rs
+++ b/argmin/src/solver/mod.rs
@@ -15,6 +15,7 @@ pub mod linesearch;
 pub mod neldermead;
 pub mod newton;
 pub mod particleswarm;
+pub mod powell;
 pub mod quasinewton;
 pub mod simulatedannealing;
 pub mod trustregion;


stefan-k:
Hi @Trombach,

thanks a lot for this PR! :) At first glance it looks already quite good. Admittedly, I'm not familiar with this method. Could you provide a reference (paper, code you were following, ...) so that I can have a look, please? 

> 1. Is my approach sensible at all? What can be improved?

I haven't seen anything that is obviously aweful, but after posting this comment I'll start a review and give some pointers where I think things could be done differently. You can also have a look at the results of the failed CI piplines, where clippy complained about a few minor things.

> 2. AFAIK, Powell's method can also be implemented using Golden-section search or Brent's method. Is there a way to do this generically in argmin or will there have to be different versions of solvers for each variant?

Do you mean as a replacement for the line search? It would be great to support both Golden-section search and Brent's method. One option would be to implement `Linesearch` for both; however I'm not sure how to implement `search_direction` for these.
Alternatively, we could consider creating an enum with `Brent(...)`, `GoldenSectionSearch(...)` and `LineSearch(...)` variants. Then you'd store this enum in your struct rather than the line search. Then I think we'd get away with separate constructors rather than different versions of the solver.

> 3. I would appreciate some suggestions on a good test function to write some tests for this solver.

I'll think about this when I'm more familiar with the method itself :)

argmin/src/solver/powell/mod.rs
```diff
@@ -0,0 +1,97 @@
+use crate::core::{
```
stefan-k, line 1: Minor: Copyright notice is missing

argmin/src/solver/powell/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+        let param = state
+            .take_param()
+            .ok_or_else(argmin_error_closure!(NotInitialized, "not initialized"))?; // TODO add Error message
+
+        // new displacement vector created from displacement vectors of line searches
+        let new_displacement = param.zero_like();
+        let mut best_direction: (usize, F) = (0, float!(0.0));
+
+        // init line search
+        let (ls_state, _) = self.linesearch.init(problem, state.clone())?;
```
stefan-k, line 55: What is the reason for initializing the line search here explicitly? This is typically called by the `Executor`.

argmin/src/solver/powell/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        // Loop over all search vectors and perform line optimization along each search direction
+        for (i, search_vector) in self.search_vectors.iter().enumerate() {
+            self.linesearch.search_direction(search_vector.clone());
+
+            let line_cost = ls_state.get_cost();
```
stefan-k, line 61: I think you're retrieving cost from `ls_state` here only to later set it again in the `configure` method of the `Executor`. It looks to me as if you only call `init` (in line 55) to evaluate the cost function. But one cannot assume that every line search will compute the cost function in the `init` method. 
I'm not exactly sure what is going on, but I believe that you can avoid the code in line 55 (where you initialize the linesearch), and also this line here and instead once compute the cost of the current parameter vector (`let cost = problem.cost(&param)?;`) prior to this loop. In line 69 you'd then pass this cost to the closure in the `configure` method.

argmin/src/solver/powell/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+        PowellLineSearch {
+            search_vectors: initial_search_vectors,
+            linesearch,
+        }
+    }
+}
+
+impl<O, P, F, L> Solver<O, IterState<P, (), (), (), F>> for PowellLineSearch<P, L>
+where
+    O: CostFunction<Param = P, Output = F>,
```
stefan-k, line 28: Since the problem is passed to a line search as well, you'll probably also require `O` to implement `Gradient`. But I'm not a 100% sure how these requirements of the LS are passed along. Maybe we just need to give it a try.

argmin/src/solver/powell/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+            self.linesearch.search_direction(search_vector.clone());
+
+            let line_cost = ls_state.get_cost();
+
+            // Run solver
+            let OptimizationResult {
+                problem: _sub_problem,
+                state: sub_state,
+                ..
+            } = Executor::new(problem.take_problem().unwrap(), self.linesearch.clone())
```
stefan-k, line 68: `take_problem` moves the actual problem out of the variable `problem`. This means that `Option::None` will be left in its place instead. This also means (apologies for this inconvenience), that after running the LS, the returned problem has to be put back into `problem` again. Otherwise the next iterations will fail because the problem is missing. In short, after line 71, you'll need a 

```rust
problem.consume_problem(sub_problem);
```

(assuming that you change `_sub_problem` to `sub_problem` in line 65). This will also take care of the function evaluation counts in `sub_problem` by adding them to the function evaluation counts in `problem`.

argmin/src/solver/powell/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+                .configure(|state| state.param(param.clone()).cost(line_cost))
+                .ctrlc(false)
+                .run()?;
+
+            // update new displacement vector
+            let displacement = &sub_state.get_best_param().unwrap().sub(&param);
+            let displacement_magnitude = displacement.dot(&displacement).sqrt();
+            new_displacement.add(displacement);
```
stefan-k, line 76: I believe this operation is not performed in-place, hence you'll need to write `new_displacement = new_displacement.add(displacement);`.
Is `new_displacement` always a vector of zeros or should it be updated in every iteration of the loop? If the latter is the case then you will require a `mut` in line 51.

argmin/src/solver/powell/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+        let _ = mem::replace(
+            &mut self.search_vectors[best_direction.0],
+            new_displacement.clone(),
+        );
+
+        // set new parameters
+        let param = param.add(&new_displacement);
+        let cost = problem.cost(&param);
+
+        Ok((state.param(param).cost(cost?), None))
```
stefan-k, line 95: This is really minor but I'd rather have the `?` in line 93 than here ;)

argmin/src/solver/powell/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    F: ArgminFloat,
+    L: Clone + LineSearch<P, F> + Solver<O, IterState<P, (), (), (), F>>,
+{
+    const NAME: &'static str = "Powell-LS";
+
+    fn next_iter(
+        &mut self,
+        problem: &mut crate::core::Problem<O>,
+        mut state: IterState<P, (), (), (), F>,
+    ) -> Result<(IterState<P, (), (), (), F>, Option<crate::core::KV>), anyhow::Error> {
```
stefan-k, line 45: It makes no difference, but for consistency I'd prefer `crate::core::Error` here instead of `anyhow::Error`.

stefan-k (changes requested):

Trombach:
Hi @stefan-k,

thanks for the feedback! Sorry for being slow to respond, but I'm doing this as a little weekend side project :) I will go through your review suggestions as soon as I can!

I have mainly followed [Wikipedia](https://en.wikipedia.org/wiki/Powell%27s_method) for now, but I will got through Powell's [paper](https://doi.org/10.1093%2Fcomjnl%2F7.2.155) as well to get a more nuanced understanding. As far as I can tell it's a gradient free method where you search along provided search directions and update your list of search directions until you converge onto a minimum.

1. Thanks, I will make sure that workflows will complete in the final PR.
2. Right, an enum might make sense here. As far as I can tell any method that can find a minimum along a provided search direction without calculating gradients should work.

stefan-k:
> thanks for the feedback! Sorry for being slow to respond, but I'm doing this as a little weekend side project :) I will go through your review suggestions as soon as I can!

There's no rush, take your time :)


> I have mainly followed [Wikipedia](https://en.wikipedia.org/wiki/Powell%27s_method) for now, but I will got through Powell's [paper](https://doi.org/10.1093%2Fcomjnl%2F7.2.155) as well to get a more nuanced understanding. As far as I can tell it's a gradient free method where you search along provided search directions and update your list of search directions until you converge onto a minimum.
> [...]
>     2. Right, an enum might make sense here. As far as I can tell any method that can find a minimum along a provided search direction without calculating gradients should work.

Thanks, I'll have a look at these. Makes sense that it works with any 1D method, but when using linesearches, it is not really a gradient free method anymore (for most line searches at least). That's fine, we just need to make that clear in the documentation :)

stefan-k:
Hi @Trombach, 

Is there any news regarding this PR? Feel free to get in touch if there is anything I can help you with :)

Trombach:
Hi @stefan-k,
 My apologies for the long silence, but other things have come up. However, I'll be working on improving this PR over the weekend. I am going to respond to some of your comments individually.
Good news: Powell published his paper with a test function so I will just be using the same for code tests.

argmin/src/solver/powell/mod.rs
Trombach: I believe one of the advantages of Powell's method is that it doesn't need gradients. I thought that as long as the linesearch method is given a direction to search along it doesn't need gradients to be computed, i.e. the calculation of gradients should be independent from the linesearch algorith, which only searches along a provided search direction. However, I could be wrong, but I can just write tests and see if it works. I will also implement a version of this using golden section search and brent's method.

argmin/src/solver/powell/mod.rs
Trombach: Thanks will add this before the PR is final. Leaving this unresolved as a reminder for myself :)

stefan-k:
> My apologies for the long silence, but other things have come up. 

No worries :)

>  Good news: Powell published his paper with a test function so I will just be using the same for code tests.

This is indeed very helpful :)

Tastaturtaste:
> Hi @stefan-k,
> 
> thanks for the feedback! Sorry for being slow to respond, but I'm doing this as a little weekend side project :) I will go through your review suggestions as soon as I can!
> 
> I have mainly followed [Wikipedia](https://en.wikipedia.org/wiki/Powell%27s_method) for now, but I will got through Powell's [paper](https://doi.org/10.1093%2Fcomjnl%2F7.2.155) as well to get a more nuanced understanding. As far as I can tell it's a gradient free method where you search along provided search directions and update your list of search directions until you converge onto a minimum.
> 
>     1. Thanks, I will make sure that workflows will complete in the final PR.
> 
>     2. Right, an enum might make sense here. As far as I can tell any method that can find a minimum along a provided search direction without calculating gradients should work.

How did you get access to the paper? I haven't found a publicly accessible version and also cannot access it through my research institution. I would like to take a look at it, too.

Tastaturtaste:
I have some remarks regarding some of the statements regarding the line search used internally.
> 2. Right, an enum might make sense here. As far as I can tell any method that can find a minimum along a provided search direction without calculating gradients should work.

That should be correct. Powell's method is a gradient free optimization algorithm, meaning the linesearch used should also not need gradient information. A typical choice according to wikipedia for such line search algorithms would be [Golden-section-search](https://en.wikipedia.org/wiki/Golden-section_search) or [Brent's method](https://en.wikipedia.org/wiki/Brent%27s_method). Another important characteristic the line search method has to fullfill is that it needs to be a bi-directional line search, meaning the coefficient of the search direction can also be negative. Considering these points maybe it would make sense to hard code a specific line search method or make a new trait `BidirectionalLinesearch` which has to be fullfilled by the use supplied line search method. Scipy has gone the way of hard coding the Brent method for the unbounded case and (as far as I can tell) the Golden-section-search for the bounded case. Currently I am working on a direct translation of scipy's implementation based on @Trombach's work.

Tastaturtaste:
> Did you manage to get hold of it? I was able to access it via my institution.

Sadly not, but I  think following the scipy implementation shouldn't be to bad for now. The algorithm can still be modified later after all.

> That's great! I don't know what @Trombach's plans are regarding this PR, but it seems as if this PR has gone stale. I'm therefore happy to consider another PR by you.

I have to see how much time I can invest at the moment, but will try my best. Unfortunately I can't compile the test suite at the moment. I get the following error in a dependency named `netlib-src`:
```
   Compiling lax v0.16.0
error: could not find native static library `blas-netlib`, perhaps an -L flag is missing?

error: could not compile `netlib-src` due to previous error
warning: build failed, waiting for other jobs to finish...
```
I don't know yet what to do to solve this issue.

Tastaturtaste:
> Unfortunately I can't compile the test suite at the moment. I get the following error in a dependency named `netlib-src`:
> 
> ```
>    Compiling lax v0.16.0
> error: could not find native static library `blas-netlib`, perhaps an -L flag is missing?
> 
> error: could not compile `netlib-src` due to previous error
> warning: build failed, waiting for other jobs to finish...
> ```
> 
> I don't know yet what to do to solve this issue.

I found a workaround for now. It seems the issue is with a ndarray-linalg feature on windows activated with this line:
```
_dev_linalg_0_16 = ["ndarray-linalg_0_16/netlib-static"]
```
The README from the ndarray-linalg repo say on windows only the Intel MKL backend is supported. Changing the cited line to 
```
_dev_linalg_0_16 = ["ndarray-linalg_0_16/intel-mkl-static"]
```
works for me. Since this backend is supported on all platforms, maybe it should be made to be the default for everyone?

stefan-k:
> Sadly not, but I think following the scipy implementation shouldn't be to bad for now. The algorithm can still be modified later after all.

Feel free to get in touch with me via stefan.kroboth AT gmail.com. I may have a solution.

> It seems the issue is with a ndarray-linalg feature on windows ...

Oh, yes, that won't work on windows. In the past, netlib was the only thing I could get to work reliably in the CI, but things may have changed for the better in the meantime. If you choose to open a PR we will see if MKL works.
Anyway apologies for the inconvenience and thanks for the fix :)

PR #244, opened 2022-08-12
Support OWL-QN method (L-BFGS with L1-regularization)

### LINKED ISSUE #243: Support OWL-QN (L-BFGS with L1 regularization)
Currently, the L-BFGS of this crate cannot perform L1-regularization because the gradient of the regularization term is the subderivative.
In the [OWL-QN](https://doi.org/10.1145/1273496.1273501) method, L1-regularization is achieved by modifying the gradient to a pseudo-gradient and performing a constrained line search.

vbkaisetsu:
Fix #243 

This PR adds the L1-regularization option to `LBFGS`.
This option enables L1-regularization performed by the OWL-QN method.


stefan-k:
Thanks a lot for this PR! I will try to review it during the weekend (but unfortunately I can't guarantee it).

At first sight I was wondering if it would be better to have a dedicated OWL-QN solver instead of adding it to the existing L-BFGS. This would lead to code duplication but it may also be easier to understand and to maintain in the long run (and maybe it would be also better in terms of performance). But I'm not sure yet about what's the best approach. I'd be interested in your opinion on this.

vbkaisetsu:
I also initially thought it would be better to define a separate structure.
However, on second thought, OWL-QN only adds a few processing to L-BFGS, and a separate structure would incur a high administrative cost since changes to L-BFGS would have to be reflected in OWL-QN every time.

argmin/src/core/test_utils.rs
```diff
@@ -207,6 +207,103 @@ impl Anneal for TestProblem {
...
+/// Example 2: x = [0, 0, 1, 1], y = -1
+/// Example 3: x = [1, 0, 0, 0], y =  1
+/// Example 4: x = [0, 0, 1, 0], y = -1
+///
+/// cost = Σ (w^T x - y)^2
+///
+/// Implements [`CostFunction`] and [`Gradient`].
+#[derive(Clone, Copy, Default, Debug, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "serde1", derive(Serialize, Deserialize))]
+
```
stefan-k, line 222: Please remove empty line.

argmin/src/solver/quasinewton/lbfgs.rs
```diff
@@ -61,6 +68,8 @@ pub struct LBFGS<L, P, G, F> {
     tol_grad: F,
     /// Tolerance for the stopping criterion based on the change of the cost stopping criterion
     tol_cost: F,
+    /// Coefficient of L1-regularization
+    l1_coeff: F,
```
stefan-k, line 72: I think I would prefer this to be `Option<F>`, where `None` represents no L1 regularization.

argmin/src/solver/quasinewton/lbfgs.rs
```diff
@@ -138,6 +148,130 @@ where
...
+    ///
+    /// ```
+    /// # use argmin::solver::quasinewton::LBFGS;
+    /// # use argmin::core::Error;
+    /// # fn main() -> Result<(), Error> {
+    /// # let linesearch = ();
+    /// let lbfgs: LBFGS<_, Vec<f64>, Vec<f64>, f64> = LBFGS::new(linesearch, 3).with_l1_regularization(1.0);
+    /// # Ok(())
+    /// # }
+    pub fn with_l1_regularization(mut self, l1_coeff: F) -> Self {
```
stefan-k, line 166: Please return an error if `l1_coeff <= 0.0`.

argmin/src/solver/quasinewton/lbfgs.rs
```diff
@@ -138,6 +148,130 @@ where
         self.tol_cost = tol_cost;
         Ok(self)
     }
+
+    /// Sets coefficient of L1-regularization.
+    ///
+    /// `0.0` or negative value disables L1-regularization. Defaults to `0.0`.
```
stefan-k, line 154: If `l1_coeff` is an `Option<F>`, I would change this to

```rust
/// Activates L1-regularization with coefficient `l1_coeff`. 
///
/// Parameter `l1_coeff` must be `> 0.0`.

File: argmin/src/solver/quasinewton/lbfgs.rs [hunk 5]
```diff
@@ -213,8 +358,17 @@ where
             float!(1.0)
         };
 
+        let mut pseudo_gradient = None;
+
         // L-BFGS two-loop recursion
-        let mut q = prev_grad.clone();
+        let mut q = if self.l1_coeff > float!(0.0) {
```
stefan-k, line 364:
```rust
let mut q = if let Some(l1_coeff) = self.l1_coeff {
```

stefan-k (changes requested):
Thanks again for this valuable contribution! I did a quick initial review with mostly minor style-related comments. Could you also provide an example `owl_qn.rs` in the `examples` folder, please? This would help me to better understand what is going on and it would of course also be useful for users.

stefan-k:
> However, on second thought, OWL-QN only adds a few processing to L-BFGS, and a separate structure would incur a high administrative cost since changes to L-BFGS would have to be reflected in OWL-QN every time.

Yes, after having a more detailed look at the code I tend to agree. The only downside is the increased number of math-related trait bounds which add an additional implementation burden on people using their own types. But I think this is acceptable for now. We can reconsider this if it starts to become a problem.

vbkaisetsu:
Thank you for your review. I fixed them.
Also, I fixed several bugs. The pseudo-gradient should also be used in line search, but the previous implementation used the normal gradient.
I added the L1-regularization version of Rosenbrock. Comparing the results with the following grid search results, I think the results are correct.

```python
# Rosenbrock w/o regularization
best = (0.0, 0.0, float('inf'))
for x in np.arange(0, 2, 0.001):
    for y in np.arange(0, 2, 0.001):
        z = (1.0 - x)**2 + 100 * (y - x**2)**2
        if z < best[2]:
            best = (x, y, z)
print(best)
#=> (1.0, 1.0, 0.0)

# Rosenbrock w/ L1 regularization (coeff = 1.0)
best = (0.0, 0.0, float('inf'))
for x in np.arange(0, 2, 0.001):
    for y in np.arange(0, 2, 0.001):
        z = (1.0 - x)**2 + 100 * (y - x**2)**2 + np.abs(x) + np.abs(y)
        if z < best[2]:
            best = (x, y, z)
print(best)
#=> (0.249, 0.057, 0.8725020001)
```

TODO: Add the L1 term to the cost function of LineSearchProblem.

argmin/src/solver/quasinewton/lbfgs.rs
```diff
@@ -138,28 +164,167 @@ where
...
+    P: ArgminMul<P, P> + ArgminMinMax + ArgminSignum + ArgminZeroLike,
+{
+    type Param = P;
+    type Output = F;
+
+    fn cost(&self, param: &Self::Param) -> Result<Self::Output, Error> {
+        if let Some(xi) = self.xi.as_ref() {
+            let zeros = param.zero_like();
+            let param = P::max(&param.mul(xi).signum(), &zeros).mul(param);
+            self.problem.cost(&param)
```
vbkaisetsu, line 256: @stefan-k This line should be as follows:
```suggestion
            self.problem.cost(&param) + self.l1_coeff.unwrap() * param.l1_norm()
```
Should I add the l1_norm() function to the `ArgminNorm` trait? Or should I create a new trait for the L1 norm?

stefan-k: I suggest adding a dedicated trait `ArgminL1Norm`. 
But then `ArgminNorm` should be renamed to `ArgminL2Norm`. If you want, you can also make that change, but it is absolutely fine if you don't want to do it because it is quite some work. I can also make that change after merging this :)

stefan-k (changes requested):
Sorry that you had to wait! I can't keep up with your impressive pace. 
Looks really good, thanks a lot for all the effort, I really appreciate it!

There is only one minor thing. Unfortunately I can't place a comment at the line that I'm talking about, so I'll have to describe it. Could you please add a small text about how this can be turned into an OWL-QN in the docs of the LBFGS struct, right above the "TODO" with the headline `## OWL-QN`? This way users will be able to more easily make the link between the L1 regularization and OWL-QN. 

After that I'd only ask you to squash your commits a bit and then I think this is ready to go! :)

vbkaisetsu:
Thank you for your review. I updated the documentation.

stefan-k (approved):
Excellent! Thanks once again for all your work, I really appreciate this great addition to the library! :)

PR #245, opened 2022-08-13
Use alias type names in examples

vbkaisetsu:
In code examples, both concrete type names and aliases are used for argument and return types of trait implementations.
This fix unifies the type name with aliases for consistency.

Also, return types of some gradient functions are incorrectly specified as `Param`, so this branch corrects them to `Gradient`.


stefan-k:
Thank you for this nice cleanup! :) Consistency is important, especially in examples.

stefan-k (approved):

PR #283, opened 2022-12-16
use as_secs_f64

TheIronBorn:


stefan-k (approved):
Much better, thanks! Must have missed that function.

PR #308, opened 2023-01-02
Check if the target cost is reached using the best cost

relf:
This PR changes slightly `terminate_internal` to use the best cost to check if the target cost is reached
instead of the current cost.

In the context of constrained optimization, the cost (the value of the cost function at the current
`param` vector) may be lower than the target cost but a constraint at this `param` vector may be violated.
In this case the target cost should not be declared to be reached.
The fix consists in using the best cost which is expected to be updated only if constraints are not violated.

### DIFF
--- a/media/book/src/implementing_solver.md
+++ b/media/book/src/implementing_solver.md
@@ -16,7 +16,7 @@ The `Solver` trait consists of several methods; however, not all of them need to
 * `init(...)`: Run before the the actual iterations and initializes the solver. Does nothing by default.
 * `next_iter(...)`: One iteration of the solver. Will be executed by the `Executor` until a stopping critereon is met.
 * `terminate(...)`: Solver specific stopping criteria. This method is run after every iteration. Note that one can also terminate from within `next_iter` if necessary.
-* `terminate_internal(...)`: By default calls `terminate` and in addition checks if the maximum number of iterations was reached or if the current cost function value is below the target cost function value. Should only be overwritten if absolutely necessary. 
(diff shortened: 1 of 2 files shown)


media/book/src/implementing_solver.md
```diff
@@ -16,7 +16,7 @@ The `Solver` trait consists of several methods; however, not all of them need to
 * `init(...)`: Run before the the actual iterations and initializes the solver. Does nothing by default.
 * `next_iter(...)`: One iteration of the solver. Will be executed by the `Executor` until a stopping critereon is met.
 * `terminate(...)`: Solver specific stopping criteria. This method is run after every iteration. Note that one can also terminate from within `next_iter` if necessary.
-* `terminate_internal(...)`: By default calls `terminate` and in addition checks if the maximum number of iterations was reached or if the current cost function value is below the target cost function value. Should only be overwritten if absolutely necessary. 
+* `terminate_internal(...)`: By default calls `terminate` and in addition checks if the maximum number of iterations was reached or if the cost function best value is below the target cost value. Should only be overwritten if absolutely necessary.
```
stefan-k, line 19: Just a minor suggestion:

```suggestion
* `terminate_internal(...)`: By default calls `terminate` and in addition checks if the maximum number of iterations was reached or if the best cost function value is below the target cost value. Should only be overwritten if absolutely necessary. 
```

stefan-k (changes requested):
Thanks a lot for this PR! Looks good to me, I only have a very small suggestion.

stefan-k (approved):

Issue #305, opened 2022-12-26
`terminate` not called on `Ctrl+C`


relf:
I've successfully argminified my solver [here](https://github.com/relf/egobox/pull/67) (thanks again for your work). 

I've used the ability to interrupt the executor using `Ctrl+C` which indeed ends the current iteration but the `terminate` method does not seem to be called at the end. As `terminate` is documented as being called after each iteration somehow I was expecting it to be also called in this case to end gracefully.  

Looking at the executor code, I was wondering if a terminaison reason `KeyboardInterrupt` could be added to handle this specific case seperately from `Aborted`. 

Maybe a better option would be to add a `final_terminate` method called only once at the end of `run` (as a symetric of `init`) to allow whatever cleanup is needed. What do you think?

stefan-k:
The purpose of `terminate` is to assess whether or not the algorithm should be stopped from iterating and feed that back to the `Executor` loop. `Ctrl+C` is a bit different however, because it directly stops the loop from progressing to the next iteration. Therefore, in case of `Ctrl+C`, the last call to the terminate function will be the one in the last loop iteration that was still executed [1].

I like the idea of a final method (which would be part of the `Solver` trait) that has mutable access to the state and the problem and which is called right before returning from `run`. But I would probably call it `finalize` instead of `final_terminate`, because it wouldn't return a `TerminationReason`. Would that help you in what you are trying to achieve?  

> Looking at the executor code, I was wondering if a terminaison reason KeyboardInterrupt could be added to handle this specific case seperately from Aborted.

Those two would then be equivalent, because `Aborted` is only used for `Ctrl+C` (so far). 

[1] I wonder if the `Ctrl+C` handling could be moved into `terminate_internal`.

relf:
Yes `finalize` is the right term.

> Would that help you in what you are trying to achieve?

At the moment my point was to only log the interruption, just to know that it was intentionally interrupted. With the `finalize` maybe I will be able to make the same check as you when you set the termination reason to `Aborted` to log the keyboard interruption. To my understanding only `Ctrl+C` is able to set the atomic bool `running` to `false`, so the `Executor` could properly handle the termination reason KeyboardInterrupt by testing this boolean after the loop. Did I miss something?

> Those two would then be equivalent, because Aborted is only used for Ctrl+C (so far).

[...]

PR #309, opened 2023-01-03
Implement KeyboardInterrupt termination

relf:
Following discussion [here](https://github.com/argmin-rs/argmin/issues/305#issuecomment-1369137557), this PR implements the `KeyboardInterrupt` termination handling.

I've tested that it works properly with my solver.

As a follow-up discussion, I do not need the `finalize` method as initially I introduced it as a way to have a hook just at the end of the loop to check the `NotTerminated` and max iter not reached to detect keyboard interrupt like the existing check at the end of the `run` function.      

Now that `KeyboardInterrupt` is handled I think this last check to set `Aborted` is useless as well, as the state is always terminated, if you're ok I remove it.

### DIFF
--- a/argmin/src/core/executor.rs
+++ b/argmin/src/core/executor.rs
@@ -267,11 +267,6 @@ where
             state = state.terminate_with(TerminationReason::KeyboardInterrupt);
         }
 
-        // In case it stopped prematurely and `termination_reason` is still `NotTerminated`,
-        // someone must have pulled the handbrake
-        if state.get_iter() < state.get_max_iters() && !state.terminated() {
-            state = state.terminate_with(TerminationReason::Aborted);
-        }
         Ok(OptimizationResult::new(self.problem, self.solver, state))
     }
 
--- a/argmin/src/core/executor.rs
+++ b/argmin/src/core/executor.rs
@@ -259,7 +259,15 @@ where
             }
         }
 
-        // in case it stopped prematurely and `termination_reason` is still `NotTerminated`,
+        if running.load(Ordering::SeqCst) {
+            // Solver execution has finished
+            running.store(false, Ordering::SeqCst);
+        } else {
+            // Solver execution has been interrupted manually
+            state = state.terminate_with(TerminationReason::KeyboardInterrupt);
+        }
+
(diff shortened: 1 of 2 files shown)


argmin/src/core/executor.rs
```diff
@@ -259,7 +259,15 @@ where
...
         }
 
-        // in case it stopped prematurely and `termination_reason` is still `NotTerminated`,
+        if running.load(Ordering::SeqCst) {
+            // Solver execution has finished
+            running.store(false, Ordering::SeqCst);
+        } else {
+            // Solver execution has been interrupted manually
+            state = state.terminate_with(TerminationReason::KeyboardInterrupt);
+        }
```
stefan-k, line 268: What is the reason for setting running to false if the solver has finished?

At first sight it seems to me that this should be sufficient:
```suggestion
        if !running.load(Ordering::SeqCst) {
            // Solver execution has been interrupted manually
            state = state.terminate_with(TerminationReason::KeyboardInterrupt);
        }
```

argmin/src/core/executor.rs
```diff
@@ -259,7 +259,15 @@ where
...
+
+        // In case it stopped prematurely and `termination_reason` is still `NotTerminated`,
         // someone must have pulled the handbrake
         if state.get_iter() < state.get_max_iters() && !state.terminated() {
             state = state.terminate_with(TerminationReason::Aborted);
```
stefan-k, line 273: I agree that your addition above makes this useless and that it can be removed.

stefan-k (changes requested):
Thanks! I only have a minor comment and I agree that the old awkward check can be removed. I'm fine with not adding `finalize` for now.

argmin/src/core/executor.rs
relf: Yep, it was my first version but somehow I was bothered to let the boolean called `running` to `true`, and yes I agree it feels awkward (even more now for there is no subsequent processing). Actually everything falls into place, if we use `interrupt` the inverse boolean of `running`.

stefan-k (approved):
Perfect, thanks!

PR #316, opened 2023-01-18
New TerminationReason definition

relf:
This PR implements a new version of `TerminationReason` enum as discussed  [here](https://github.com/argmin-rs/argmin/issues/305#issuecomment-1382015565).

It introduces the variant `SolverExit(String)` which allows to define custom stopping reason and `SolverConverged` which is intended to gather all successful exit reasons under the same umbrella. 

Some unwanted implementation consequences: 
* `TerminationReason` is not `Copy` anymore (`TerminationStatus` as well)
* We "loose" information when several successful exits are possible for a solver


stefan-k (approved):
Thanks a lot, I really appreciate all the work!!

PR #320, opened 2023-01-20
Mention egobox in docs and Readme

stefan-k:
@relf I'd like to mention egobox in the Readme and the docs in the hope that others will follow your example ;) Please let me know if you're for whatever reason not ok with that.


relf:
Awesome! Thanks! 😃

Issue #278, opened 2022-12-16
Replace `assert!` macros with macros from approx crate


stefan-k:
Currently most assertions in the argmin-math tests use the `assert!` macro, which is somewhat inconvenient for comparing floats. It would be great if those were replaced with the assert macros of the [approx](https://docs.rs/approx/latest/approx/) crate.

Status 

* [x] primitives (#327)
* [x] vec (#327)
* [x] ndarray (#404)
* [x] nalgebra (#404)

rpopplewell:
So for the sake of clarity we're looking for this sort of change for all tests hit by `cargo test -p argmin-math` right?
```                 
// Old
// assert!(((a as $t - res.re) as f64).abs() < std::f64::EPSILON);
// assert!(((-b as $t - res.im) as f64).abs() < std::f64::EPSILON);

// New
assert_relative_eq!(a as f64, res.re as f64, epsilon = std::f64::EPSILON);
assert_relative_eq!(-b as f64, res.im as f64, epsilon = std::f64::EPSILON);
```

stefan-k:
Yes, this is correct! :)

PR #326, opened 2023-01-24
Replace assert! macros with macros from approx crate

rpopplewell:
(https://github.com/argmin-rs/argmin/issues/278) Replaces all `assert!` macros that directly compare `f64s` in tests. `approx::assert_relative_eq(f64::NAN, f64::NAN)` does not pass, so I skipped those.

### DIFF
--- a/argmin/src/solver/conjugategradient/cg.rs
+++ b/argmin/src/solver/conjugategradient/cg.rs
@@ -182,8 +182,8 @@ mod tests {
             p_prev,
             rtr,
         } = cg;
-        assert_relative_eq!(b[0], 1.0f64);
-        assert_relative_eq!(b[1], 2.0f64);
+        assert_relative_eq!(b[0], 1.0f64, epsilon = f64::EPSILON);
+        assert_relative_eq!(b[1], 2.0f64, epsilon = f64::EPSILON);
         assert!(r.is_none());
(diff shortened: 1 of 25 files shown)


stefan-k:
Thanks for the huge effort! I haven't done a thorough review yet, but I can already say one thing: Whenever I used `.to_ne_bytes()`, I did it deliberately, because I want to make sure that they are *exactly* the same. Using `to_ne_bytes()` better communicates the intention; however, I probably should have also made a comment about this somewhere.

rpopplewell:
Ah my bad, I assumed I'd continue the pattern that was there for existing approx comparisons. Let me know what to change though (besides the rust fmt of course)

stefan-k:
I just had a bit more detailed look at it and I'm afraid that most of the work you have done is not needed and this is entirely my fault. I was not precise enough in the issue you linked (#278). Actually, what I meant was the test code in argmin-math and not in argmin. I should have specified that, I am very sorry!! 
 
I really appreciate the effort you put in, and if you are still interested in doing the changes to argmin-math then I'm happy to review another PR.

rpopplewell:
Ah I should've asked I suppose. You can close this and I'll see to getting around to it later. It's a cool project and it helps me learn rust, so no worries!

PR #340, opened 2023-03-19
PoC: Python bindings

jjbayer:
This is a PoC for adding Python bindings to argmin. See PR comments for open questions.

Currently, only the Newton solver is implemented.

TODO:

- [ ] Call provided cost functions efficiently.
- [ ] Add more solvers.
- [ ] Add observers.
- [ ] Documentation.
- [ ] Python tests.


argmin/src/core/executor.rs
```diff
@@ -26,7 +26,7 @@ pub struct Executor<O, S, I> {
     /// Storage for observers
     observers: Observers<I>,
     /// Checkpoint
-    checkpoint: Option<Box<dyn Checkpoint<S, I>>>,
+    checkpoint: Option<Box<dyn Checkpoint<S, I> + Send>>,
```
jjbayer, line 29: I added these bounds because `PyClass` must be `Send`.

argmin-py/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+name = "argmin"
+crate-type = ["cdylib"]
+
+[dependencies]
+anyhow = "1.0.70"
+argmin_testfunctions = "0.1.1"
+argmin = {path="../argmin", default-features=false, features=[]}
+argmin-math = {path="../argmin-math", features=["ndarray_latest-serde"]}
+ndarray-linalg = { version = "0.16", features = ["netlib"] }
+ndarray = { version = "0.15", features = ["serde-1"] }
```
jjbayer, line 16: This section requires cleanup, I am not sure what's the best configuration of features.

argmin-py/src/problem.rs
```diff
@@ -0,0 +1,68 @@
...
+    InputDimension: ndarray::Dimension,
+    OutputDimension: ndarray::Dimension,
+{
+    // TODO: prevent dynamic dispatch for every call
+    Python::with_gil(|py| {
+        let args = PyTuple::new(py, [param.to_pyarray(py)]);
+        let pyresult = callable.call(py, args, Default::default())?;
+        let pyarray = pyresult.extract::<&numpy::PyArray<Scalar, OutputDimension>>(py)?;
+        // TODO: try to get ownership instead of cloning
+        Ok(pyarray.to_owned_array())
```
jjbayer, line 66: 1. I am unsure what's the overhead of calling `to_pyarray` and `extract` for every evaluation of the gradient, hessian etc. Probably needs benchmarks.
2. [`to_owned_array`](https://docs.rs/numpy/latest/numpy/array/struct.PyArray.html#method.to_owned_array) makes a copy of the data, this should not be necessary.

jjbayer:
Requires https://github.com/argmin-rs/argmin/pull/341.

argmin-py/Cargo.toml
stefan-k: This looks mostly fine. I think in the long run at least the `serde1` feature of argmin can be enabled because that would allow checkpointing. But I guess checkpointing will need more work anyways.
At some point we will have to decide which BLAS backend to use. This is probably mostly a platform issue (only Intel-MKL works on Linux, Windows and Mac) and a licensing issue since the compiled code will be packaged into a python module.

argmin-py/src/executor.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[pyclass]
+pub struct Executor(Option<core::Executor<Problem, DynamicSolver, IterState>>);
+
+impl Executor {
+    /// Consumes the inner executor.
+    ///
+    /// PyObjects do not allow methods that consume the object itself, so this is a workaround
+    /// for using methods like `configure` and `run`.
+    fn take(&mut self) -> anyhow::Result<core::Executor<Problem, DynamicSolver, IterState>> {
```
stefan-k, line 26: Nice! I see the point for `run`, but I don't really see it for `configure`. In case of `run` the `Executor` is consumed, but in the python example, `configure` is used as if it were using `&mut self`. I am probably missing something.

argmin-py/src/executor.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn configure(&mut self, kwargs: Option<&PyDict>) -> PyResult<()> {
+        if let Some(kwargs) = kwargs {
+            let param = kwargs
+                .get_item("param")
+                .map(|x| x.extract::<&PyArray1>())
+                .map_or(Ok(None), |r| r.map(Some))?;
+            let max_iters = kwargs
+                .get_item("max_iters")
+                .map(|x| x.extract())
+                .map_or(Ok(None), |r| r.map(Some))?;
```
stefan-k, line 51: I really like how you transformed the somewhat weird way one needs to set the initial state in argmin (using a closure) into a very pythonic `**kwargs` thing. This may however turn into quite a chore given that there are multiple kinds of state (`IterState` and `PopulationState`) with lots of methods. However, I would label this problem low priority for now.

argmin-py/src/problem.rs
stefan-k: I just had a new idea: You're currently using the `ndarray` backend which requires transitioning between numpy arrays and ndarray arrays. Instead we could add a new math backend based on `PyArray`, which would mean that numpy would do all the heavy lifting. I'm not sure whether numpy or ndarray is faster and I haven't really thought this through either.

I somehow thought that it would be possible to use the underlying memory in both Rust and Python without copying but that doesn't seem to be the case. 

Regarding point 2: I agree. I assumed that there is also `into_owned_array` but that does not seem to be the case.

argmin-py/src/solver.rs
```diff
@@ -0,0 +1,51 @@
...
+};
+
+#[pyclass]
+#[derive(Clone)]
+pub enum Solver {
+    Newton,
+}
+
+pub enum DynamicSolver {
+    // NOTE: I tried using a Box<dyn Solver<> here, but Solver is not object safe.
```
stefan-k, line 26: Too bad!! That's what I was hoping for, but I tend to forget about object safety. I don't think it'll be possible to make it object safe :(

argmin-py/src/solver.rs
```diff
@@ -0,0 +1,51 @@
...
+impl From<Solver> for DynamicSolver {
+    fn from(solver: Solver) -> Self {
+        match solver {
+            Solver::Newton => Self::Newton(solver::newton::Newton::new()),
+        }
+    }
+}
+
+impl core::Solver<Problem, IterState> for DynamicSolver {
+    // TODO: make this a trait method so we can return a dynamic
```
stefan-k, line 39: Sounds good to me! We could have both for backwards compatibility, right? The default impl of the `name` method would then just return `self.NAME`.

argmin-py/src/types.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2018-2023 argmin developers
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+//! Base types for the Python extension.
+
+pub type Scalar = f64; // TODO: allow complex numbers
```
stefan-k, line 10: Complex numbers would be great but I wouldn't give this a high priority.

argmin/src/core/executor.rs
stefan-k: Sounds also good to me. Would it make sense to add the `Send` bound to the `Checkpoint` trait?

argmin-py/src/problem.rs
jjbayer: > I somehow thought that it would be possible to use the underlying memory in both Rust and Python without copying but that doesn't seem to be the case.

I'll give it a try!

argmin-py/src/solver.rs
jjbayer: It seems I was able to solve two problems at once: When I remove the associated constant, `Solver` becomes object-safe, so we can create trait objects for it. Let me know if you objects.

media/book/src/implementing_solver.md
```diff
@@ -1,7 +1,7 @@
 # Implementing a solver
 
 In this section we are going to implement the Landweber solver, which essentially is a special form of gradient descent.
-In iteration \\( k \\), the new parameter vector \\( x_{k+1} \\) is calculated from the previous parameter vector \\( x_k \\) and the gradient at \\( x_k \\) according to the following update rule:
+In iteration \\( k \\), the new parameter vector \\( x\_{k+1} \\) is calculated from the previous parameter vector \\( x_k \\) and the gradient at \\( x_k \\) according to the following update rule:
```
jjbayer, line 4: My autoformatter made most changes to this file.

argmin-py/src/executor.rs
stefan-k: Thanks for the explanation!

argmin-py/src/solver.rs
stefan-k: Nice! I thought the generics would also be a problem for object safety, but it's great if this isn't the case. Sounds good to me!

media/book/src/implementing_solver.md
stefan-k: I wasn't sure if the change in line 4 would render correctly but surprisingly it does.

jjbayer:
Closing as stale.

PR #341, opened 2023-03-19
Nightly lints

jjbayer:
Fix warnings generated by nightly clippy.

### DIFF
--- a/argmin/src/solver/quasinewton/lbfgs.rs
+++ b/argmin/src/solver/quasinewton/lbfgs.rs
@@ -402,6 +402,7 @@ where
         };
 
         // L-BFGS two-loop recursion
+        #[allow(clippy::redundant_clone)]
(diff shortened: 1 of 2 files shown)


argmin/src/solver/quasinewton/lbfgs.rs
```diff
@@ -402,6 +402,7 @@ where
         };
 
         // L-BFGS two-loop recursion
+        #[allow(clippy::redundant_clone)]
```
jjbayer, line 405: This is not really a redundant clone, because line 428 needs the unmodified previous gradient.

stefan-k: Interesting that clippy doesn't catch that.

stefan-k (approved):
LGTM! Thanks a lot!

Issue #91, opened 2021-01-17
Allow users to create stopping criteria


stefan-k:
Currently, stopping criteria are defined during the implementation of the solver. However, sometimes it might be useful if users could add their own stopping criteria. This would require a complete redesign of how termination works. One of the key aspects of termination is that it needs access to the fields of the solver which makes it difficult to implement in a generic way. However, similar to how observers work, the needed variables could be stored in a key-value store and passed to the user-defined termination function. It needs to be evaluated how efficient such an approach would be.

KSchmidAmilar:
We've had success with writing our own execution loop based on the code in Executor.run(). This gives us more fine-grained control over termination criteria.   This approach is similar to the api for libgsl.  Perhaps eliminating the Executor and allowing the user to write the loop would achieve this customizability.

stefan-k:
I was planning to approach this the same way as it is done for observers. A list of (user-defined) stopping criteria (defined by a trait) can be attached to an `Executor`, which will all be evaluated in each iteration. A stopping criteria has access to the current state (and maybe the key-value metrics which are passed to the observers). The question remains how to get access to the fields of the solver, but maybe this isn't really needed. 

These stopping criteria would be in addition to the stopping criteria implemented in the respective solver.

Would you require more flexibility than that?

PR #346, opened 2023-05-04
Add trait f64 f32 for 1d inverse to ArgminInv

sdrap:
I tried to implement the f64/f32 standard inverse for the ArgminInv (modification in the implementation of inv for ndarray).
I also added a test for it.

For unknown compilation reasons on my machine (admittedly weird linux conf), I could not run the test nor use the library locally to check if that works. I will have to look later into it.

I am also not sure if this is what was needed (as a fully newcomer into rust).

**Short Update:** I could run the optimization for scalar function, it seems to work. I am not sure how to get the resulting `get_best_param` or `get_param` in the end, it seems that this function changed in 0.8.

### DIFF
--- a/argmin-math/src/ndarray_m/inv.rs
+++ b/argmin-math/src/ndarray_m/inv.rs
@@ -22,6 +22,14 @@ macro_rules! make_inv {
                 Ok(<Self as Inverse>::inv(&self)?)
             }
         }
+
+        // inverse for scalars (1d solvers)
+        impl ArgminInv<$t> for $t {
+            #[inline]
+            fn inv(&self) -> Result<$t, Error> {
+                Ok(1.0 / self)
+            }
+        }
     };
 }
 
@@ -58,6 +66,16 @@ mod tests {
                             assert!((((res[(i, j)] - target[(i, j)]) as f64).abs()) < 0.000001);
                         }
                     }
+                }
+            }
+
+            item! {
+                #[test]
+                fn [<test_inv_scalar_ $t>]() {
+                    let a = 2.0;
+                    let target = 0.5;
+                    let res = <$t as ArgminInv<$t>>::inv(&a).unwrap();
+                    assert!(((res - target) as f64).abs() < 0.000001);
                 }
             }
         };


stefan-k:
> I tried to implement the f64/f32 standard inverse for the ArgminInv (modification in the implementation of inv for ndarray). I also added a test for it.

Thanks a lot!! :)
 
> For unknown compilation reasons on my machine (admittedly weird linux conf), I could not run the test nor use the library locally to check if that works. I will have to look later into it.

Have you seen [this section of the book](https://www.argmin-rs.org/book/contributing.html)? The CI passes, so it should be fine.

> **Short Update:** I could run the optimization for scalar function, it seems to work. I am not sure how to get the resulting `get_best_param` or `get_param` in the end, it seems that this function changed in 0.8.

You should be able to get an `Option<&f64>` via `result.state().get_best_param()`. Alternatively you should be able to access the `best_param` field of the state with `result.state().best_param`. *(Yes, this should have better ergonomics ;))*

sdrap:
Thanks, I could get it after a couple of try -> `*res.state().best_param.as_ref().unwrap()`.

BTW: I was wondering how to get a termination done with newton when for instance $f^\prime'/f^{\prime\prime}' is below a certain tolerance level.

I am also thinking about getting a 1d version where the optimum should remain within an interval (for instance when the function is undefined (+infty) out of the bounds. But for that maybe I just implement a special solver since it is arguably quite special.

Many thanks in any cases, the library works like a charm.

stefan-k:
> I was wondering how to get a termination done with newton when for instance $f^\prime'/f^{\prime\prime}' is below a certain tolerance level.

Unfortunately argmin doesn't yet allow users to create their own stopping criteria (#91). You could however add this stopping criterion to the solver and issue a PR, similarly to [how it is done here](https://github.com/argmin-rs/argmin/blob/0bcc00c45596dd35ed3b92d587fb85bc7a09e89a/argmin/src/solver/quasinewton/lbfgs.rs#L496). 

> I am also thinking about getting a 1d version where the optimum should remain within an interval (for instance when the function is undefined (+infty) out of the bounds. But for that maybe I just implement a special solver since it is arguably quite special.

argmin unfortunately doesn't have a proper concept of bounds yet (apart from Simulated Annealing and Particle Swarm Optimization, where it is quite straight-forward). You can implement this in your own code by implementing the [`Solver`](https://docs.rs/argmin/latest/argmin/core/trait.Solver.html) trait. That way you are able to run the solver in the `Executor` with all its benefits. Feel free to get in touch if you run into problems with this, I'm happy to get some feedback about the ergonomics of implementing a solver. 

> Many thanks in any cases, the library works like a charm.

Thanks! :)

stefan-k (approved):
Thanks a lot! :)

PR #349, opened 2023-05-09
SlogLogger emit data config

itrumper:
I found myself wanting to be able to customize what data the `SlogLogger` observer was logging, so I attempted to implement this functionality via the builder pattern as shown in the doc example in the `data` method on `SlogLogger`.

To do this, I needed to add the `Debug` trait as a trait bound to the `type Param` associated types, hence why this PR touches so many files. This could cause breaking changes for users.

Since this uses a builder style pattern to define what data is logged, the default contains the same data as what was previously hard coded, so end users should not experience any breaking changes in their observer configurations. To customize the data that is logged, a new `enum` called `StateData` has been defined in the `state/mod.rs` file that covers all the available data you can get from the `State` trait. A user supplies a `Vec` of these enum variants and then the logger iterates through and emits them. The `Debug` trait was needed because I wanted to print out the `Param`s as well, which are generics.

The motivation for doing this was to actually be able to log the gradient of a problem as well, but since not every solver has a defined gradient, I'm less sure that can be done. I will need assistance in evaluating that possibility, since it would significantly help with debugging!

Let me know what you think! Thanks for building a cool crate :)

### DIFF
--- a/argmin/src/core/executor.rs
+++ b/argmin/src/core/executor.rs
@@ -572,6 +572,8 @@ mod tests {
     #[test]
     #[cfg(feature = "serde1")]
     fn test_checkpointing_solver_initialization() {
+        use std::fmt::Debug;
+
         use crate::core::checkpointing::{CheckpointingFrequency, FileCheckpoint};
         use crate::core::test_utils::TestProblem;
         use crate::core::{ArgminFloat, CostFunction};
@@ -588,7 +590,7 @@ mod tests {
         impl<O, P, F> Solver<O, IterState<P, (), (), (), F>> for OptimizationAlgorithm
         where
             O: CostFunction<Param = P, Output = F>,
-            P: Clone,
+            P: Clone + Debug,
             F: ArgminFloat,
         {
             const NAME: &'static str = "OptimizationAlgorithm";
--- a/argmin/src/core/mod.rs
+++ b/argmin/src/core/mod.rs
@@ -53,5 +53,5 @@ pub use problem::{CostFunction, Gradient, Hessian, Jacobian, LinearProgram, Oper
 pub use result::OptimizationResult;
 pub use serialization::{DeserializeOwnedAlias, SerializeAlias};
 pub use solver::Solver;
-pub use state::{IterState, LinearProgramState, PopulationState, State};
+pub use state::{IterState, LinearProgramState, PopulationState, State, StateData};
 pub use termination::{TerminationReason, TerminationStatus};
--- a/argmin/src/core/observers/slog_logger.rs
+++ b/argmin/src/core/observers/slog_logger.rs
@@ -13,6 +13,7 @@
 //! See [`SlogLogger`] for details regarding usage.
 
 use crate::core::observers::Observe;
+use crate::core::state::StateData;
 use crate::core::{Error, State, KV};
 use slog;
 use slog::{info, o, Drain, Key, Record, Serializer};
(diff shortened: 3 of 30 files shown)


stefan-k:
Thanks a lot for this work! I apologize again to you for the very late response. Unfortunately you chose the worst possible time to issue a PR this big as I'm currently extraordinarily busy ;) Therefore I'm unfortunately unable to give a detailed review right know, but I do want to give you a bit of feedback before you invest more time into this PR.

Generally, I'm very happy to see filtering implemented in the SlogLogger; however, I'm not convinced that this approach is the way to go. A few points:

* Having an enum to represent the possible metrics is limiting, because each solver can define their own metrics sent to the observer interface. In order to be as flexible as possible, I think there is no way to get past string comparisons.
* A `Debug` bound on `Param` shouldn't be necessary (I think). I would have to look at it in detail, but the observer itself could enforce this bound without it being used all over the place in the core code. As an example you could have a look at file checkpointing, where a `Serialize` bound is only required when checkpointing is actually used. I'm however not sure how well this also applies to the observer case.
* I don't think state is the appropriate place to store which fields to observe. As this is observer specific, I think it should be possible to only change the observer code. 

I had the need for filtering also in a new observer I am working on in #311. I chose to collect a list of strings and then only select the ones listed (see line 148 in [this file](https://github.com/argmin-rs/argmin/pull/311/files#diff-9487794018acf665944cd11d5a445fc49692c9162e7e09ef6b7a3b58491c4bfe)). Note that this code essentially does something quite different to what you want to achieve, but I think the approach should be similar.

Regarding observing parameters/gradients and so on: I would love to see this, but maybe it would be easier to implement this in a dedicated observer? I'm not sure though, just an idea to think about :)

Thanks again for this contribution, I really appreciate it! I hope that not too much of your work needs to be changed and I hope that I will be able to give you a detailed review soon, but unfortunately I cannot promise it right now.

itrumper:
Thanks for taking the time to look it over and provide feedback! That is a great help, and thanks for letting me know about your availability. I don't have expectations of a timeline, so please don't feel pressured to work on this.

In response to your points:

1. I see that point, we need a flexible enough interface to accommodate `Solver`s different parameters. However, I want to try to find a way that implements better type safety than string comparisons, but still is extensible. I don't know what that looks like right now, so I'll think about it more.
2. I will certainly look into the file checkpointing! Seems like you're describing a trait bound behind a feature gate?
3. I may have miscommunicated about this point. Which data to log isn't stored in `State` but rather the `Observer` itself. I only modified that file to add the `enum` that defines what "names" of data are available.

I'll check out the observer you are working on, thanks for the references!

For the solver specific data, like parameters and gradients, I do think it would take a different type of observer, as right now the `Observer` doesn't know which parameters are available, just what is exposed through the `State` trait, correct?

I'm happy to completely redo it all if a better solution is out there! You know your crate, and how pieces fit together better than I do, so I'll happily take your direction. Thanks for the discussion.

stefan-k:
>     1. I see that point, we need a flexible enough interface to accommodate `Solver`s different parameters. However, I want to try to find a way that implements better type safety than string comparisons, but still is extensible. I don't know what that looks like right now, so I'll think about it more.

I agree that string comparisons aren't the most elegant solution, not only for type safety, but also for performance reasons.

>     2. I will certainly look into the file checkpointing! Seems like you're describing a trait bound behind a feature gate?

`FileCheckpoint` is behind the `serde` feature gate, but this is not what's "(de)activating" the trait bound. `FileCheckpoint` imposes a `Serialize` trait bound on the solver and the state ([here](https://github.com/argmin-rs/argmin/blob/13a39ea20855ebaa5747caac47ce0177b3ff8a1d/argmin/src/core/checkpointing/file.rs#L80)), but only in case `FileCheckpoint` is actually used. If a user doesn't use checkpointing in their code, the state and solver do not need to be serializable. 

>     3. I may have miscommunicated about this point. Which data to log isn't stored in `State` but rather the `Observer` itself. I only modified that file to add the `enum` that defines what "names" of data are available.

I see, sorry for the misunderstanding! 

> For the solver specific data, like parameters and gradients, I do think it would take a different type of observer, as right now the `Observer` doesn't know which parameters are available, just what is exposed through the `State` trait, correct?

That's a good point! The `State` trait does expose the parameters via `get_param` and `get_best_param` ([here](https://github.com/argmin-rs/argmin/blob/13a39ea20855ebaa5747caac47ce0177b3ff8a1d/argmin/src/core/state/mod.rs#L62-L66)). Gradients are not accessible via the `State` trait, but it should be possible to create a dedicated `GradientObserver` which only works with `IterState` states. Then you will have access to all fields of the `IterState` and it is easy to turn the observer on and off for debugging.

PR #354, opened 2023-05-11
Correct line search to search over gradient

DevonMorris:
Previously, the line search operated directly over parameter space. This caused issues when there were different types for parameters vs gradients e.g. Lie Groups. This commit changes line searches to operate over gradient parameterization and incorporates gradient updates to parameters via ArgminScaledAdd.

I'm not sure sure what level of testing is needed for this but it passes all the unit tests with `cargo test` and I also ran the 3 line search examples and the steepest descent example

### DIFF
--- a/argmin/examples/steepestdescent_manifold.rs
+++ b/argmin/examples/steepestdescent_manifold.rs
@@ -1,0 +1,94 @@
+// Copyright 2018-2022 argmin developers
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+#![allow(unused_imports)]
+
+use argmin::core::observers::{ObserverMode, SlogLogger};
+use argmin::core::{CostFunction, Error, Executor, Gradient};
+use argmin::solver::gradientdescent::SteepestDescent;
+use argmin::solver::linesearch::condition::{ArmijoCondition, LineSearchCondition};
+use argmin::solver::linesearch::BacktrackingLineSearch;
+use argmin_math::ArgminScaledAdd;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Copy, Debug)]
+struct ClosestPointOnCircle {
+    x: f64,
+    y: f64,
+}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
+struct CirclePoint {
+    angle: f64,
+}
+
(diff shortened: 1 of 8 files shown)


argmin/examples/steepestdescent_manifold.rs
```diff
@@ -0,0 +1,95 @@
...
+impl ArgminScaledAdd<f64, f64, CirclePoint> for CirclePoint {
+    fn scaled_add(&self, alpha: &f64, delta: &f64) -> Self {
+        CirclePoint {
+            angle: self.angle + alpha * delta,
+        }
+    }
+}
+
+fn run() -> Result<(), Error> {
+    // Define cost function (must implement `ArgminOperator`)
```
stefan-k, line 63:
```suggestion
    // Define cost function (must implement `CostFunction` and `Gradient`)
```

Admittedly, this is also wrong in the "original" steepestdescent.rs ;)

stefan-k (changes requested):
Thanks a lot for this contribution! This is a case I didn't consider. I have only a minor request in the example you provided (which is actually my mistake). 

Another thing that I would appreciate if you could do is to change `P` in the definition of the `LineSearch` trait to `D` ([here](https://github.com/argmin-rs/argmin/blob/2457232e27a1fc0fb0b95f77eebaa939d34aea93/argmin/src/solver/linesearch/mod.rs#L77)). This is obviously purely cosmetic but I think it would make it less tempting to use `P` again.

argmin/examples/steepestdescent_manifold.rs
DevonMorris: Everything here should be resolved. I also added some fixups for the CI failures. Let me know once you've reviewed and I can squash those.

argmin/src/solver/linesearch/mod.rs
```diff
@@ -55,16 +55,16 @@ pub use self::morethuente::MoreThuenteLineSearch;
 /// ```
 /// use argmin::solver::linesearch::LineSearch;
 ///
-/// struct MyLineSearch<P, F> {
-///     // `P` is the type of the search direction, typically the same as the parameter vector
-///     search_direction: P,
+/// struct MyLineSearch<D, F> {
+///     // `D` is the type of the search direction, typically the same as the parameter vector
```
stefan-k, line 59: Thanks for also fixing the docs. There is only one small suggestions:
```suggestion
///     // `D` is the type of the search direction
```

stefan-k (changes requested):
Thanks! There is only one minor comment. Once this is fixed this is ready to be squashed and merged :)

argmin/src/solver/linesearch/mod.rs
DevonMorris: Should be resolved now. I also squashed all the fixup commits.

stefan-k (approved):
Thanks! :)

PR #359, opened 2023-05-27
Add a simple example of Nelder-Mead usage.

cjordan:
Hi,

Thanks for the effort that's been put into `argmin`. I've recently used the Nelder-Mead part successfully and it will be a great addition to my tool belt.

I felt the need to contribute a simple example, because over the last couple of years, I've tried a couple of times to use `argmin`, and the existing docs and examples were not clear enough for me.  I wanted to use Nelder-Mead, but I (even still) didn't know what a Rosenbrock function is in the existing example (Wikipedia suggests that this is a sensible thing to test optimisation algs, but my lack of familiarity was intimidating).  This PR contains another Nelder-Mead example, which demonstrates searching for a cubic polynomial's roots.  It also shows how to extract the optimised parameter from the `Executor` result, which was not obvious to me from the existing docs or examples.

Please feel free to edit what I've provided, and I'm open to suggestions on making this better.  My objective is to make what already is a well-polished library even more accessible, and allow people like myself to continue developing in Rust instead of having to write more code to access e.g. scipy optimize instead.

### DIFF
--- a/argmin/Cargo.toml
+++ b/argmin/Cargo.toml
@@ -130,7 +130,7 @@ name = "morethuente"
 required-features = ["slog-logger"]
 
 [[example]]
-name = "neldermead-simple"
+name = "neldermead-cubic"
 required-features = ["slog-logger"]
 
 [[example]]
--- a/argmin/Cargo.toml
+++ b/argmin/Cargo.toml
@@ -130,6 +130,10 @@ name = "morethuente"
 required-features = ["slog-logger"]
 
 [[example]]
+name = "neldermead-simple"
+required-features = ["slog-logger"]
+
+[[example]]
 name = "neldermead"
 required-features = ["argmin-math/ndarray_latest-serde", "slog-logger"]
 
--- a/argmin/examples/neldermead-simple.rs
+++ b/argmin/examples/neldermead-simple.rs
@@ -1,0 +1,129 @@
+// Copyright 2018-2022 argmin developers
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+//! A (hopefully) simple example of using Nelder-Mead to find the roots of a
+//! cubic polynomial.
+//!
+//! You can run this example with:
+//! `cargo run --example neldermead-simple --features slog-logger`
+
+use argmin::core::observers::{ObserverMode, SlogLogger};
+use argmin::core::{CostFunction, Error, Executor, State};
+use argmin::solver::neldermead::NelderMead;
+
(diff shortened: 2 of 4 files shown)


stefan-k:
Hi @cjordan, thanks a lot for this PR!! Improvements of documentation and examples are very welcome, and I'm very thankful for the effort you put into this! :) Having good documentation to make argmin more accessible is very important to me, but it is also a very time-cosuming task and as such every help is highly appreciated! 

I apologize for the very very late reply, unfortunately I've been rather busy over the past months and currently only have limited access to the internet (#360). 

Right now I have time for a rough review and I hope I'll be able to give a more in-depth review at a later stage (if even necessary).

argmin/Cargo.toml
```diff
@@ -129,6 +129,10 @@ required-features = ["argmin-math/nalgebra_latest-serde", "slog-logger"]
 name = "morethuente"
 required-features = ["slog-logger"]
 
+[[example]]
+name = "neldermead-simple"
```
stefan-k, line 133: I think I would prefer `neldermead-cubic`.

argmin/examples/neldermead-simple.rs
```diff
@@ -0,0 +1,129 @@
...
+            .configure(|state| state.max_iters(100))
+            .add_observer(SlogLogger::term(), ObserverMode::Always)
+            .run()?;
+
+        // Wait a second (lets the logger flush everything before printing again)
+        std::thread::sleep(std::time::Duration::from_secs(1));
+
+        // Print result
+        println!("{res}");
+        // Print our result in a custom way.
```
stefan-k, line 83: I think these two lines can be removed (people who have not looked at the other examples may be confused).

argmin/src/solver/neldermead/mod.rs
```diff
@@ -12,6 +12,9 @@
 //!
 //! See [`NelderMead`] for details.
 //!
+//! A simple example of usage can be found at
+//! <https://github.com/argmin-rs/argmin/blob/main/argmin/examples/neldermead-simple.rs>
+//!
```
stefan-k, line 17: In theory, I am all in favor of advertising the example here. The problem however is, that this points to the current main branch. If people use older versions of argmin (and the corresponding docs) they may be led to an example which doesn't work for them. We have had this in the past, and decided to remove all links to examples in the docs. One could link to the corresponding example via the git tags; however, it is a major chore to update all links for every new version.

stefan-k (changes requested):
Thanks again for this very useful addition! I only have a couple of minor comments.

cjordan:
Hi @stefan-k, thanks for reviewing my PR, and I'm glad you find it useful. No worries about being late to reply; I understand the feeling of being busy!

 I agree with all of your points, and I've now made a new commit. Feel free to squash the commits (as I like to do to appease my OCD), or let me know if you want me to do it.

stefan-k (approved):
Thanks for you work and your patience! :) This looks good to me! :)

PR #386, opened 2024-01-14
Remove cargo-deny exception for intel-mkl-src

Tastaturtaste:
As conversed about in PR #369, this PR removes the cargo-deny exclusion for intel-mkl-src introduced due to https://github.com/EmbarkStudios/krates/issues/60. 

This PR should be merged as soon as https://github.com/EmbarkStudios/krates/pull/61 lands in cargo-deny-action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: argmin CI
 
 on:
+  workflow_dispatch:
   pull_request:
   push:
     branches:
@@ -207,6 +208,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - uses: EmbarkStudios/cargo-deny-action@v1
+    - uses: EmbarkStudios/cargo-deny-action@v2
       with:
         command: check ${{ matrix.checks }}
--- a/deny.toml
+++ b/deny.toml
@@ -17,6 +17,11 @@
 # this list would mean the nix crate, as well as any of its exclusive
 # dependencies not shared by any other crates, would be ignored, as the target
 # list here is effectively saying which targets you are building for.
+
+# We only care about the licenses of dependencies we have to respect with our own license
+# So crates that are not published or that only appear in dev dependencies can be ignored
+exclude-unpublished = true
+exclude-dev = true
 targets = [
     # The triple can be any string, but only the target triples built in to
     # rustc (as of 1.40) can be checked against actual config expressions
--- a/deny.toml
+++ b/deny.toml
@@ -8,7 +8,7 @@
 
 # The values provided in this template are the default values that will be used
 # when any section or field is not specified in your own configuration
-
+[graph]
 # If 1 or more target triples (and optionally, target_features) are specified,
 # only the specified targets will be checked when running `cargo deny check`.
 # This means, if a particular package is only ever used as a target specific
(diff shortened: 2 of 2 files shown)


jedbrown:
It seems this is all complete. Should this be merged now?

Tastaturtaste:
> It seems this is all complete. Should this be merged now?

The CI build should be rerun to confirm. If that is green it should be good to merge.

stefan-k:
Would you mind rebasing this PR or allowing me to push to it? Thanks! :)

Tastaturtaste:
I can rebase in the next few days. Also, permission for maintainer to edit this PR should already be given.

Tastaturtaste:
Seems like cargo-deny picks up intel-mkl-src through the "examples" crates, even though those crates have "publish = false" set in their manifest and "deny.toml" has "licenses.private.ignore = true" set. If the "examples" folder is just removed from the workspace manifest members there are no license errors. I think this might not be the intended behavior for the "licenses.private.ignore" option of `cargo deny`. Maybe this is another bug in their implementation? I don't have the time to look closer into this right now. Maybe in the next few days, but no promises.

stefan-k:
Thanks for the quick rebase, and thanks for digging into this. I've also had a look, and I suspect the path based dependencies are treated differently. I tried to replace them by workspace dependencies but to no avail. If you do look into this, please do not waste too much time on this annoying problem.

> Also, permission for maintainer to edit this PR should already be given.

Interestingly I still cannot push to this branch.

Tastaturtaste:
I get no errors with `cargo deny --all-features --log-level warn check licenses bans sources` in a local checkout of my branch, both on windows and linux (wsl) with cargo-deny@v0.16.2. Seems to me there is something wrong with the cargo-deny-action? Seems the right version of cargo-deny is used in the workflow, but the necessary keys `graph.exclude-unpublished` and `graph.exclude-dev` are not used in the call to cargo-deny. I opened https://github.com/EmbarkStudios/cargo-deny-action/issues/81 to investigate.

Tastaturtaste:
Turns out I was wrong to suspect the cargo-deny-action. I just didn't see errors locally since I did run `cargo update` and thus my cargo.lock was different from the one generated with a fresh checkout of the branch in the workflow. It appears with updates to some crates the "Unicode-3.0" license got into the dependency graph. 

According to https://www.unicode.org/policies/licensing_policy.html the license is based on the MIT license, with the difference that the Unicode-3.0 license is targeted explicitly at data and data files:

> The Consortium’s software and data files are generally licensed under the [OSI-approved Unicode License v3,](https://opensource.org/license/unicode-license-v3) a free, open source, highly permissive license based on the [MIT License](https://opensource.org/license/MIT). The primary difference between the MIT License and the Unicode License is that the Unicode License expressly covers data and data files. The Unicode License v3 and its predecessors have been approved by the Open Source Initiative (OSI) since 2015.

It seems the license is very permissive. However to be sure I would follow suite with the most restrictive use of the license in all of the dependencies, which appears to be https://crates.io/crates/unicode-ident, and change the license of argmin to "(MIT OR Apache-2.0) AND Unicode-3.0". Or track down why dependencies with this license got into the dependency graph in the first place and try to prune them.

stefan-k:
A huge thank you for the work on this issue and your endurance, this is remarkable! :)

Tastaturtaste:
> A huge thank you for the work on this issue and your endurance, this is remarkable! :)

Thank you for these kind words. I learned a lot along the way!

PR #391, opened 2024-01-17
Test cases for `ArgminMinMax`

Shreyan11:
Please do let me know if there's anything else to be done,I'm open to all suggestions.


README.md
stefan-k, line 1: There are still changes to this file which should not be there.

argmin-math/src/primitives/minmax.rs
```diff
@@ -49,3 +49,27 @@ make_minmax!(Complex<isize>);
 make_minmax!(Complex<usize>);
 
 // TODO: tests!!!
```
stefan-k, line 51: Please remove this line.

stefan-k (changes requested):
Thank you! Please don't open a new PR for each review, you can do all changes in one PR. There are still unnecessary changes to `README.md`, probably due to a failed rebase.

argmin-math/src/primitives/minmax.rs
Shreyan11: I've updated it,it should be good,let me know if there's anything else to be done

stefan-k (approved):
Looks good to me! Thanks!

PR #479, opened 2024-03-01
Integrate finitediff into argmin monorepo

stefan-k:
- Started integrating finitediff into argmin monorepo
- Added CI pipeline for finitediff, fixed doc test
- Updated finitediff Cargo.toml
- Fixed ndarray warning in finitediff


crates/finitediff/src/vec/jacobian.rs
```diff
@@ -0,0 +1,270 @@
...
+                .collect::<Vec<f64>>()
+        })
+        .collect()
+}
+
+pub fn central_jacobian_vec_f64(x: &[f64], fs: &dyn Fn(&Vec<f64>) -> Vec<f64>) -> Vec<Vec<f64>> {
+    let mut xt = x.to_owned();
+    (0..x.len())
+        .map(|i| {
+            let fx1 = mod_and_calc(&mut xt, fs, i, EPS_F64.sqrt());
```
jedbrown, line 30: For central differencing, you're better off using the cube root so you can improve accuracy to $\epsilon_{\text{machine}}^{2/3}$. With the square root, you'd still only get $\epsilon_{\text{machine}}^{1/2}$ error in the derivative.
```suggestion
            let fx1 = mod_and_calc(&mut xt, fs, i, EPS_F64.cbrt());
```
If you choose sqrt, you get the intersection of the red and magenta lines.
image

That said, it's dubious to rely on the problem being carefully nondimensionalized in a library implementaion of finite differencing. The two algorithms I know of are from [Walker and Pernice](https://petsc.org/release/manualpages/Mat/MATMFFD_WP/) and from [Dennis and Schnabel](https://petsc.org/release/manualpages/Mat/MATMFFD_DS/). In my experience, some problems benefit significantly from the more expensive (Dennis and Schnabel) method.

stefan-k: Thanks for pointing that out! I followed the Nocedal and Wright book on this (p. 169), and after reading that section again it's clear that I must have missed that part. If I understand it correctly,  the book states it should be `EPS_F64.cbrt().powi(2)`, and not `EPS_F64.cbrt()`. What am I missing?

Thanks for pointing me to the other resources which I wasn't aware of. I will have a look, but probably not as part of this PR.

jedbrown: Hah, indeed. That's a mistake on page 169 (it was intended to state that as the achievable error), but it refers to the exercises, which has the correct statement (as in my comment and figure above). 

image

In analogy to Equation 7.5, where you want to choose $\epsilon$ to minimize $\epsilon + \symbf u/\epsilon$ (truncation error plus rounding error), when using central differences the truncation error is $O(\epsilon^2)$ so you wish to minimize $\epsilon^2 + \symbf u/\epsilon$, leading to $\epsilon = \symbf u^{1/3}$. (I dropped the $L$ and $L_f$ used to reach (7.6), which scalings like WP or DS seek to respect.)

stefan-k: Thanks for clarifying and for the explanation! Interestingly, this mistake isn't even mentioned in the [errata](http://users.iems.northwestern.edu/~nocedal/book/2ndprint.pdf). I will change it to `eps.cbrt()` then :)

Issue #478, opened 2024-02-29
Thoughts on future state handling


stefan-k:
State is needed to keep track of any kind of data from one iteration to the next. This state could be handled by the solver structs themselves, but by using a dedicated state, it is possible to define a common interface and it avoids having to re-implement certain functionality for every solver. However, I'm not really happy with the current design of `IterState` and `PopulationState`. `IterState` is designed to accommodate a wide range of solvers, but in doing so it is neither very space efficient, nor is it convenient to use due to its many internal type parameters.
If a solver only operates on parameters and gradients, it still carries `Option<Jacobian>` and others around. Some solvers are sufficiently different to justify their own kind of state, such as `ParticleSwarm`, which has `PopulationState`. 

Ideally, each solver should have it's own state with a yet-to-be-designed common interface. However, I'm afraid that implementing this will be an annoying and unnecessary seeming chore. Also, the common interface must not suffer from similar problems as the current design.

My current idea is to provide "state building blocks" which can be stitched together using a declarative macro. For instance, there could be dedicated building blocks for handling parameter vectors, gradients, Hessians, Jacobians, residuals, populations, number of function calls, costs, time, termination status, ... probably each with their own Trait, which are automatically implemented for the "aggregated" state type. For each solver, one just has to define which building blocks are needed and the macro will take care of it all. I'm not sure though if all of this can be done with a declarative macro. 

One potential problem I see with this is in all the places which depend on the state implementing the `State` trait. 

This design should also allow one to reuse state from one run to another, as this could avoid unnecessary allocations. This would require some sort of resetting mechanism which each of the building blocks could provide. 

Any ideas on this are highly welcome!

PR #483, opened 2024-03-12
Some minor optimizations

### LINKED ISSUE #482: Various non productive overhead
I noticed that `argmin` collects statistics about how many times functions are called, for example here and in a few other places

https://github.com/argmin-rs/argmin/blob/35e00e2bdfe2ebd83f6c1ee29d83abd60edbdcc0/crates/argmin/src/core/state/iterstate.rs#L1358-L1363

Both allocations and hashing can be expensive compared to solving, in fact I hacked `examples/newton` to skip console logging and to run the computation a bunch more times taking care of blackboxing the results - counting is responsible for about 20% of all the time.

And if I disable timing (`Executor::timer`) along with removing counting - newton example is suddenly 30% faster...

Any objections if I try to shove it out of the way? Timing doesn't seem to be used unless there are observers - should probably be disabled unless needed, counting can probably be made optional or made into observer, not 100% sure yet.

pacak:
- Add a benchmark (that works on stable toolchain) to establish the baseline
- Disable timing by default, enable when needed
- Disable execution counting by default

after:
```
newton  fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ run  4.9 µs        │ 171.4 µs      │ 5.25 µs       │ 6.218 µs      │ 4723907 │ 4723907
```

before:
```
    newton  fastest       │ slowest       │ median        │ mean          │ samples │ iters
    ╰─ run  6.317 µs      │ 171.8 µs      │ 6.729 µs      │ 7.75 µs       │ 3807698 │ 3807698
```

Fixes #482

### DIFF
--- a/crates/argmin/src/core/executor.rs
+++ b/crates/argmin/src/core/executor.rs
@@ -69,7 +69,7 @@ where
             checkpoint: None,
             timeout: None,
             ctrlc: true,
-            timer: true,
+            timer: false,
         }
     }
 
@@ -383,9 +383,9 @@ where
         self
     }
 
-    /// Enables or disables timing of individual iterations (default: enabled).
+    /// Enables or disables timing of individual iterations (default: false).
     ///
-    /// Setting this to false will silently be ignored in case a timeout is set.
+    /// In case a timeout is set, this will automatically be set to true.
     ///
     /// # Example
     ///
@@ -768,7 +768,7 @@ mod tests {
         let problem = TestProblem::new();
         let timeout = std::time::Duration::from_secs(2);
 
-        let executor = Executor::new(problem, solver);
+        let executor = Executor::new(problem, solver).timer(true);
         assert!(executor.timer);
         assert!(executor.timeout.is_none());
 
--- a/crates/argmin/src/core/state/iterstate.rs
+++ b/crates/argmin/src/core/state/iterstate.rs
@@ -971,6 +971,20 @@ where
     pub fn take_prev_residuals(&mut self) -> Option<R> {
         self.prev_residuals.take()
     }
+
+    /// Overrides state of counting function executions (default: false)
+    /// ```
+    /// # use argmin::core::{IterState, State};
+    /// # let mut state: IterState<(), (), (), (), Vec<f64>, f64> = IterState::new();
+    /// # assert!(!state.counting_enabled);
+    /// let state = state.counting(true);
+    /// # assert!(state.counting_enabled);
+    /// ```
+    #[must_use]
+    pub fn counting(mut self, mode: bool) -> Self {
+        self.counting_enabled = mode;
+        self
+    }
 }
 
 impl<P, G, J, H, R, F> State for IterState<P, G, J, H, R, F>
--- a/crates/argmin/src/core/state/iterstate.rs
+++ b/crates/argmin/src/core/state/iterstate.rs
@@ -81,6 +81,8 @@ pub struct IterState<P, G, J, H, R, F> {
     pub max_iters: u64,
     /// Evaluation counts
     pub counts: HashMap<String, u64>,
+    /// Update evaluation counts?
+    pub counting_enabled: bool,
     /// Time required so far
     pub time: Option<instant::Duration>,
     /// Status of optimization execution
@@ -1039,6 +1041,7 @@ where
             last_best_iter: 0,
             max_iters: std::u64::MAX,
             counts: HashMap::new(),
+            counting_enabled: false,
             time: Some(instant::Duration::new(0, 0)),
             termination_status: TerminationStatus::NotTerminated,
         }
@@ -1338,7 +1341,7 @@ where
     /// ```
     /// # use std::collections::HashMap;
     /// # use argmin::core::{Problem, IterState, State, ArgminFloat};
-    /// # let mut state: IterState<Vec<f64>, (), (), (), (), f64> = IterState::new();
+    /// # let mut state: IterState<Vec<f64>, (), (), (), (), f64> = IterState::new().counting(true);
     /// # assert_eq!(state.counts, HashMap::new());
     /// # state.counts.insert("test2".to_string(), 10u64);
     /// #
@@ -1355,9 +1358,11 @@ where
     /// # assert_eq!(state.counts, hm);
     /// ```
     fn func_counts<O>(&mut self, problem: &Problem<O>) {
-        for (k, &v) in problem.counts.iter() {
-            let count = self.counts.entry(k.to_string()).or_insert(0);
-            *count = v
+        if self.counting_enabled {
+            for (k, &v) in problem.counts.iter() {
+                let count = self.counts.entry(k.to_string()).or_insert(0);
+                *count = v
+            }
         }
     }
 
--- a/crates/argmin/src/core/state/linearprogramstate.rs
+++ b/crates/argmin/src/core/state/linearprogramstate.rs
@@ -56,6 +56,8 @@ pub struct LinearProgramState<P, F> {
     pub max_iters: u64,
     /// Evaluation counts
     pub counts: HashMap<String, u64>,
+    /// Update evaluation counts?
+    pub counting_enabled: bool,
     /// Time required so far
     pub time: Option<instant::Duration>,
     /// Status of optimization execution
@@ -150,6 +152,20 @@ impl<P, F> LinearProgramState<P, F> {
         self.cost = cost;
         self
     }
+
+    /// Overrides state of counting function executions (default: false)
+    /// ```
+    /// # use argmin::core::{State, LinearProgramState};
+    /// # let mut state: LinearProgramState<Vec<f64>, f64> = LinearProgramState::new();
+    /// # assert!(!state.counting_enabled);
+    /// let state = state.counting(true);
+    /// # assert!(state.counting_enabled);
+    /// ```
+    #[must_use]
+    pub fn counting(mut self, mode: bool) -> Self {
+        self.counting_enabled = mode;
+        self
+    }
 }
 
 impl<P, F> State for LinearProgramState<P, F>
@@ -205,6 +221,7 @@ where
             last_best_iter: 0,
             max_iters: std::u64::MAX,
             counts: HashMap::new(),
+            counting_enabled: false,
             time: Some(instant::Duration::new(0, 0)),
             termination_status: TerminationStatus::NotTerminated,
         }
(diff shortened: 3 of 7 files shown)


pacak:
Commit that adds benchmark is not necessary, I can drop it or change to use criterion.

pacak:
I'll add some docs in a second, but I'm not sure what the failures are about.

crates/argmin/src/core/executor.rs
```diff
@@ -383,7 +386,7 @@ where
         self
     }
 
-    /// Enables or disables timing of individual iterations (default: enabled).
+    /// Enables or disables timing of individual iterations (default: when needed).
     ///
     /// Setting this to false will silently be ignored in case a timeout is set.
```
stefan-k, line 391:
```suggestion
    /// Enables or disables timing of individual iterations (default: false).
    ///
    /// In case a timeout is set, this will automatically be set to true.
```

Now that I think of it, I think this should probably be set to `false` by default (independent of whether observers are used or not), as it makes documentation substantially easier.

crates/argmin/src/core/state/mod.rs
```diff
@@ -90,6 +90,9 @@ pub trait State {
     /// Returns current cost function evaluation count
     fn get_func_counts(&self) -> &HashMap<String, u64>;
 
+    /// Set function counting status
+    fn set_counting(self, mode: bool) -> Self;
+
```
stefan-k, line 95: I don't like this being part of the `State` trait, as it leaks too much internal details in a public interface. I'd have to have a more detailed look though in order to be able to make suggestion.
Also, setters are usually not called `set_*` in argmin, therefore I'd prefer `counting`.

crates/argmin/src/core/state/iterstate.rs
```diff
@@ -1356,9 +1359,11 @@ where
     /// # assert_eq!(state.counts, hm);
     /// ```
     fn func_counts<O>(&mut self, problem: &Problem<O>) {
-        for (k, &v) in problem.counts.iter() {
-            let count = self.counts.entry(k.to_string()).or_insert(0);
-            *count = v
+        if self.counting_enabled {
+            for (k, &v) in problem.counts.iter() {
+                let count = self.counts.entry(k.to_string()).or_insert(0);
```
stefan-k, line 1364: By the way, have you been able to find out whether hashing is the problem or the allocation of `.to_string()`? In case it was the latter, I was wondering if `Cow` would help?
If it's hashing, I was wondering if replacing `HashMap<String, u64>` with `HashMap<SomeEnum, u64>` would help, where 

```rust
enum SomeEnum {
    CostCount,
    GradientCount,
    OperatorCount,
    JacobianCount,
    HessianCount,
    AnnealCount,
    Other(String),
}
```

(or something along those lines). That would cover all the function calls that in argmin, but also allows one to count function calls defined in external code via `Other`. It also avoids `.to_string()`.

stefan-k (changes requested):
Thank a lot for your work! I have a few comments. I understand why you made `set_counting` part of the `State` trait, but I'd rather not expose this functionality that way. Is it possible to make it part of the interface of the various states, but not the `State` trait? 
This essentially highlights some of the issues I have with the current design of state handling (#478). I will have to think about it, but that may take a bit.

Btw, have you tried whether making line 224 in `executor.rs` optional helps? Unfortunately I can't make a review comment there because the PR doesn't touch this code. I mean this line:

```rust
state.func_counts(&self.problem);
```

I understand that it does not remove all counts handling, but I'm just interested how that affects performance. 

Finally I would prefer to not have the benchmark as part of this PR, but you can of course keep it in the PR until it is ready to be merged.

crates/argmin/src/core/state/iterstate.rs
pacak: As far as I remember it's 35% allocation and 65% hashing. `HashMap` is DoS resistant by default. Changing the hash function or switching to `BTreeMap` combined with enum approach you propose might help a bit I guess, but this changes public API.

pacak:
> Is it possible to make it part of the interface of the various states, but not the State trait?

Should be possible for as long as I don't add "enabled if observers are present". Since you suggest to remove that logic - I can move it to part of the each interface. Will try that today.

> Btw, have you tried whether making line 224 in executor.rs optional helps?
> I understand that it does not remove all counts handling, but I'm just interested how that affects performance.

This removes about half of the overhead related to counting I think. Better than nothing 

> Finally I would prefer to not have the benchmark as part of this PR, but you can of course keep it in the PR until it is ready to be merged.

Absolutely, I just wanted to show what exactly I'm measuring. Will drop before merging.

pacak:
1. made a change to comments
2. renamed `set_counting` to `counting` and moved it to corresponding states instead of a trait
3. removed logic that enables counting when observer is present

crates/argmin/src/core/executor.rs
```diff
@@ -313,6 +313,7 @@ where
         mode: ObserverMode,
     ) -> Self {
         self.observers.push(observer, mode);
+        self.timer = true;
```
stefan-k, line 316: Apologies, I probably miscommunicated this: I'd prefer it if timer was off by default, even if an observer is added. The only case where `timer` should be set to `true` by default is when a timeout is set.

crates/argmin/src/core/state/iterstate.rs
stefan-k: I don't mind changing the public API in that case. Why do you propose a `BTreeMap` instead of a `HashMap`? Hashing an enum should be quick (I expect it to use it's integer representation as a hash with a bit of additional overhead for hashing `Other(String)`). Anyway I think this goes beyond this PR and can certainly be left as a future improvement.

stefan-k (changes requested):
Apologies for the very late reply, unfortunately I wasn't able to get to it earlier.

crates/argmin/src/core/executor.rs
pacak: Should be fixed.

pacak:
> Hashing an enum should be quick (I expect it to use it's integer representation as a hash with a bit of additional overhead for hashing Other(String)).

By default `HashMap` uses SipHash 1-3 with random seed which is slow. What's more underlying SwissTable hash doesn't perform well when values are not random. Either way - I don't mind which implementation is used as long as it can be disabled. To minimize the overhead of running with it enabled if changing the API is not a problem I'd look into [`enum_map`](https://docs.rs/enum-map/latest/enum_map/) crate. `SomeEnum` you proposed above can then be parametrized with some type `T` for custom measurements by external users. `enum_map` maps values into an array so should be pretty fast. But yea, this is outside of the scope of this pull request.

pacak:
New changes: 

1. I removed the benchmark
2. Adding an observer will not enable timing

pacak:
Fixed the failing test. FWIW I'm getting test failures around the places I never touched...


This one is related to floating point accuracy - probably EPSILON is too large..
```
$ cargo test -p argmin --all-features

---- solver::gaussnewton::gaussnewton_linesearch::tests::test_next_iter_regression stdout ----
thread 'solver::gaussnewton::gaussnewton_linesearch::tests::test_next_iter_regression' panicked at crates/argmin/src/solver/gaussnewton/gaussnewton_linesearch.rs:445:9:
assert_relative_eq!(state.param.as_ref().unwrap()[1], 2.25f64, epsilon = f64::EPSILON)

    left  = 2.2499999999999964
    right = 2.25


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    solver::gaussnewton::gaussnewton_linesearch::tests::test_next_iter_regression
```

Not sure what is this about at all
```
$ cargo test --all-features
   Compiling argmin-math v0.4.0 (/home/pacak/ej/argmin/crates/argmin-math)
error[E0277]: the trait bound `num_complex::Complex<f32>: ndarray_linalg::Lapack` is not satisfied
  --> crates/argmin-math/src/ndarray_m/inv.rs:18:13
   |
18 |             Array2<$t>: Inverse,
   |             ^^^^^^^^^^^^^^^^^^^ the trait `ndarray_linalg::Lapack` is not implemented for `num_complex::Complex<f32>`
...
38 | make_inv!(Complex<f32>);
   | ----------------------- in this macro invocation
   |
   = help: the following other types implement trait `ndarray_linalg::Lapack`:
             f32
             f64
             nalgebra::Complex<f32>
             nalgebra::Complex<f64>
   = note: required for `ArrayBase<OwnedRepr<num_complex::Complex<f32>>, ndarray::Dim<[usize; 2]>>` to implement `Inverse`
   = help: see issue #48214
   = note: this error originates in the macro `make_inv` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `num_complex::Complex<f64>: ndarray_linalg::Lapack` is not satisfied
  --> crates/argmin-math/src/ndarray_m/inv.rs:18:13
   |
18 |             Array2<$t>: Inverse,
   |             ^^^^^^^^^^^^^^^^^^^ the trait `ndarray_linalg::Lapack` is not implemented for `num_complex::Complex<f64>`
...
39 | make_inv!(Complex<f64>);
   | ----------------------- in this macro invocation
   |
   = help: the following other types implement trait `ndarray_linalg::Lapack`:
             f32
             f64
             nalgebra::Complex<f32>
             nalgebra::Complex<f64>
   = note: required for `ArrayBase<OwnedRepr<num_complex::Complex<f64>>, ndarray::Dim<[usize; 2]>>` to implement `Inverse`
   = help: see issue #48214
   = note: this error originates in the macro `make_inv` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `argmin-math` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: could not compile `argmin-math` (lib test) due to 2 previous errors
```

stefan-k:
> Fixed the failing test. FWIW I'm getting test failures around the places I never touched...

I suspect you need to rebase onto main. There was a Rust version update and that typically causes new clippy lints to fail, but that was fixed in #488 .

> This one is related to floating point accuracy - probably EPSILON is too large..
> 
> ```
> $ cargo test -p argmin --all-features
> ...
> ```

We've had this before -- it typically only fails locally, not in the CI. We couldn't figure out why ... if it does fail in the CI too I'm happy to increase the epsilon, but judging from the current state of the CI run, it seems to pass.

> Not sure what is this about at all

That's really strange... but it seems to pass in the CI.

I hope to be able to do another review today! :)

pacak:
> I suspect you need to rebase onto main.

Done. In fact I have scripts to automagically rebase onto origin/master, then someone smart decided to rename the default branch.

stefan-k (approved):
LGTM, thanks a lot!

PR #487, opened 2024-03-24
Make `Solver` object safe

jjbayer:
Replace associated constant for solver names with a method to make `Solver` object safe.

> A trait is object safe if it has the following qualities (defined in [RFC 255](https://github.com/rust-lang/rfcs/blob/master/text/0255-object-safety.md)):
> 
>    - [...]
>    - It must not have any associated constants.

https://doc.rust-lang.org/beta/reference/items/traits.html#object-safety


jjbayer:
Requires https://github.com/argmin-rs/argmin/pull/488 to pass CI.

stefan-k (approved):
Thanks a lot!

PR #508, opened 2024-08-21
Make the LineSearch gradient parameter in BFGS consistent with the impl

dspyz-matician:
I want to implement my own custom `Param` type, but use `Array1<f32>` as the gradient type. However, the current constraints essentially force these types to be the same because the `BacktrackingLineSearch` trait is only implemented for <G, F> not for <P, F>

This PR fixes BFGS's constraints so that G is used consistently as the parameter to `BFGS`'s associated `LineSearch`


stefan-k (approved):
Thank you! :)

Issue #403, opened 2024-01-20
BrentRoot is missing tests


stefan-k:
[`BrentRoot`](https://github.com/argmin-rs/argmin/blob/main/argmin/src/solver/brent/brentroot.rs) is not tested at all. Adding these tests could be a great starting point for someone who wants to start contributing.

PR #525, opened 2024-09-30
Add tests to BrentRoot

GermanHeim:
Hi, this is my PR for #403.

I added an error message for when a negative tolerance is passed as an argument.
I created a couple of tests:
- Negative tolerance
- Invalid range when $f(a)$ and $f(b)$ have the same sign
- Valid range
- Check that the algorithm converges for a quadratic problem
- Symmetry (check that if passing $a$ as $b$ gives the same result)

I implemented a quadratic problem ($x^2 - 1$) to do the tests, but I am unsure if this is the correct way to do it in the library.
Let me know if it needs to be changed.

### DIFF
--- a/crates/argmin/src/solver/brent/brentroot.rs
+++ b/crates/argmin/src/solver/brent/brentroot.rs
@@ -18,6 +18,9 @@ pub enum BrentRootError {
     /// f(min) and f(max) must have different signs
     #[error("BrentRoot error: f(min) and f(max) must have different signs.")]
     WrongSign,
+    // tol must be positive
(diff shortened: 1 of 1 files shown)


stefan-k:
Oh dear, I wanted to rebase this PR onto the new main to fix the clippy issue and somehow completely messed up and directly pushed your changes to main! Not really sure how this can happen since the main branch is protected... 
Apparently I'm not doing this enough recently! 

Luckily I don't have any complaints with this PR! :) Thanks a lot for your work on this and apologies for messing up the merge!

PR #530, opened 2024-10-21
LaTeX support for argmin-testfunctions docs

### LINKED ISSUE #418: Add LaTeX support to docs
In particular `argmin-testfunctions` could profit from that. There are a few discussions on how that may be done:

* https://github.com/rust-lang/rust/issues/34261
* https://stackoverflow.com/questions/46495063/how-to-write-math-formulas-for-rust-documentation/54573800#54573800

airwoodix:
This patch adds LaTeX/[KaTeX](https://katex.org/) support to the `argmin-testfunctions` crate documentation.

The HTML code in `argmin-testfunctions/katex-header.html` is added to the header of all the documentation pages in that crate and loads the KaTeX [auto-render](https://katex.org/docs/autorender.html) extension from the recommended CDN. It may be possible to serve the KaTeX code from local files but I didn't investigate this.

As an example, the documentation of a couple of test function is migrated to using KaTeX rendering. For example that of the `ackley` function:

image

If it is acceptable, the migration of other documentation pages shall be done in subsequent PRs.

As a fly-by, I added some missing cross links and fixed some typos. This patch doesn't contain any Rust code changes.

Resolves #418.

### Local build

The workspace's top-level Cargo configuration sets the [`RUSTDOCFLAGS`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#:~:text=RUSTDOCFLAGS) to [include the HTML header](https://doc.rust-lang.org/rustdoc/command-line-arguments.html#--html-in-header-include-more-html-in-head) in all docs pages. This enables KaTeX support for the documentation of *all* crates for local builds. This is misleading because there's no support in the `docs.rs` build for crates other than `argmin-testfunctions` (see below). Unfortunately, [Cargo does not read config files from crates within the workspace](https://doc.rust-lang.org/cargo/reference/config.html#:~:text=Cargo%20does%20not%20read%20config%20files%20from%20crates%20within%20the%20workspace) so there's no easy way to restrict that configuration to a single crate while keeping the unified workspace build.

### `docs.rs` build

The `docs.rs` build was tested on a `docs.rs` development build. The `docs.rs` build is configured through data in the `package.metadata.docs.rs` table in the crate's manifest. Workspace metadata [is not supported](https://github.com/rust-lang/docs.rs/issues/2226). Also, the referenced files must be part of the crate, so enabling support for other crates would probably require duplicating the `katex-header.html` file to the other crates, which is a bit of a maintenance burden.

### Error handling

Syntax errors in the KaTeX code unfortunately only lead to broken rendering. Since the KaTeX code is interpreted in the browser, it may be quite involved to automatically propagate errors to the `rustdoc` call.

### Reference

https://github.com/victe/rust-latex-doc-minimal-example (and many other publicly available crates)

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustdocflags = [ "--html-in-header", "./crates/argmin-testfunctions/katex-header.html" ]
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -33,3 +33,6 @@ jobs:
       - name: Test code samples
         working-directory: ./media/book/tests
         run: cargo test
+        env:
+          # Override value in workspace top-level Cargo config.toml
+          RUSTDOCFLAGS: ''
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,9 @@ jobs:
 
   tests-argmin-math:
     runs-on: ubuntu-latest
+    env:
+      # Override value in workspace top-level Cargo config.toml
+      RUSTDOCFLAGS: ''
     steps:
       - uses: actions/checkout@v4
       - uses: Swatinem/rust-cache@v2
--- a/crates/argmin-testfunctions/Cargo.toml
+++ b/crates/argmin-testfunctions/Cargo.toml
@@ -12,6 +12,9 @@ readme = "README.md"
 keywords = ["test", "function", "optimization"]
 categories = ["science"]
 
+[package.metadata.docs.rs]
+rustdoc-args = [ "--html-in-header", "katex-header.html" ]
+
 [dependencies]
 num = "0.4"
 
--- a/crates/argmin-testfunctions/katex-header.html
+++ b/crates/argmin-testfunctions/katex-header.html
@@ -1,0 +1,13 @@
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.11/dist/katex.min.css" integrity="sha384-nB0miv6/jRmo5UMMR1wu3Gz6NLsoTkbqJghGIsx//Rlm+ZU03BU6SQNC66uf4l5+" crossorigin="anonymous">
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.11/dist/katex.min.js" integrity="sha384-7zkQWkzuo3B5mTepMUcHkMB5jZaolc2xDwL6VFqjFALcbeS9Ggm/Yr2r3Dy4lfFg" crossorigin="anonymous"></script>
+<script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.11/dist/contrib/auto-render.min.js" integrity="sha384-43gviWU0YVjaDtb/GhzOouOXtZMP/7XUzwPTstBeZFe/+rCMvRwr4yROQP43s0Xk" crossorigin="anonymous"></script>
+<script>
+    document.addEventListener("DOMContentLoaded", function() {
+        renderMathInElement(document.body, {
+            delimiters: [
+                {left: "$$", right: "$$", display: true},
+                {left: "$", right: "$", display: false},
+            ]
+        });
+    });
+</script>
--- a/crates/argmin-testfunctions/src/ackley.rs
+++ b/crates/argmin-testfunctions/src/ackley.rs
@@ -9,12 +9,14 @@
 //!
 //! Defined as
 //!
-//! `f(x_1, x_2, ..., x_n) = - a * exp( -b \sqrt{\frac{1}{d}\sum_{i=1}^n x_i^2 ) -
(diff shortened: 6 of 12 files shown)


airwoodix:
Setting `build.rustdocflags` in the workspace root's Cargo configuration file breaks the book doctests workflow, because the path to `katex-header.html` doesn't resolve from `media/book/tests`. As a workaround, I added an override in the workflow step, such that the `RUSTDOCFLAGS` are empty when running the book doctests.

There's a similar issue with a similar workaround for the `argmin-math` tests (and anything that is not invoked with the workspace root as working directory).

Open point: is it preferable to set `RUSTDOCFLAGS='--html-in-header ...'` only during the actual docs build (instead of the global setting in `.cargo/config.toml`)?

stefan-k:
Firstly, this is an absolutely magnificent PR, both in terms of work done and description! Thank you for the massive effort you have taken to make this work!  

Do I understand it correctly that the main problem is that both `docs.rs` and local builds have to be configured/set up separately? If that is the case, I think I can live with it.

> Open point: is it preferable to set RUSTDOCFLAGS='--html-in-header ...' only during the actual docs build (instead of the global setting in .cargo/config.toml)?
Would this fix the discrepancy between `docs.rs` and local builds, or would this avoid the work arounds for `argmin-math` and `media/book/tests` (and instead requires setting `RUSTDOCSFLAGS` for `argmin-testfunctions`)?

The failing clippy lints should be fixed in main.

airwoodix:
Thanks for the feedback!

> Do I understand it correctly that the main problem is that both docs.rs and local builds have to be configured/set up separately?

Yes, that's correct.

> Would this fix the discrepancy between docs.rs and local builds, or would this avoid the work arounds for argmin-math and media/book/tests (and instead requires setting RUSTDOCSFLAGS for argmin-testfunctions)?

It would remove the workarounds for `media/book/tests` and `argmin-math`. Setting `RUSTDOCFLAGS` for the CI step would work and remove the new `.cargo/config.toml` and the mentioned workarounds:

```diff
diff --git a/.github/workflows/docs.yml b/.github/workflows/docs.yml
index bbd3a70c..affe0c1f 100644
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -19,6 +19,8 @@ jobs:
           components: rustfmt,rust-src

       - name: Build documentation
+        env:
+          RUSTDOCFLAGS: "--html-in-header ./crates/argmin-testfunctions/katex-header.html"
         run: cargo doc --features "argmin-math/latest_all,argmin/full" --no-deps --workspace --exclude "example-*"

       - name: Forwarding
```

Caveats:

- building locally during development is more cumbersome (one also needs to set `RUSTDOCFLAGS` in the local environment)
- the HTML header is again applied to all the crates, not just to `argmin-testfunctions`. I'm not aware of a way to achieve the latter while keeping the single doc build command issued from the workspace root.

> The failing clippy lints should be fixed in main.

Thanks! I rebased the branch on `main`.

stefan-k:
Apologies once again for the long wait! 


> > Do I understand it correctly that the main problem is that both docs.rs and local builds have to be configured/set up separately?
> 
> Yes, that's correct.

This is fine with me.

> Caveats:
> 
>     * building locally during development is more cumbersome (one also needs to set `RUSTDOCFLAGS` in the local environment)
> 
>     * the HTML header is again applied to all the crates, not just to `argmin-testfunctions`. I'm not aware of a way to achieve the latter while keeping the single doc build command issued from the workspace root.

Caveat 2 seems fine with me, but caveat 1 does sound cumbersome and it would be nice if this could be avoided. 

Unfortunately I lost track a bit of this PR (apologies again, because I can't really express how excited I am about this!). Are there any decisions you need from me, or can this be merged?

airwoodix:
> caveat 1 does sound cumbersome

I agree. The options I currently see are either an opinionated way of setting the environment (e.g. [direnv](https://direnv.net/)) or introducing a task runner (e.g. [just](https://github.com/casey/just) or `make`), both or which would be quite a change.

> Are there any decisions you need from me, or can this be merged?

I rebased the branch and would stick with the initial version:

- locally, `cargo doc --no-deps` works from the workspace root, at the expense of the not-so-nice workarounds for `book` and `argmin-math` in the CI.
- building the `argmin-testfunctions` crate by docs.rs works too.

stefan-k:
> I rebased the branch and would stick with the initial (current) version:
> 
>     * locally, `cargo doc --no-deps` works from the workspace root, at the expense of the not-so-nice workarounds for `book` and `argmin-math` in the CI.
> 
>     * building the `argmin-testfunctions` crate by docs.rs works too.

Thanks, this sounds very good to me! Thanks for rebasing. Unfortunately the CI was still broken. Could you do another rebase, please? It should hopefully be fixed now :)

airwoodix:
@stefan-k Thanks for the feedback and the main branch fix. I rebased again and all CI jobs pass now on this branch too.

stefan-k (approved):

PR #544, opened 2024-12-01
New Root Solver: ITP Method

duncanam:
# Intro

Hello! I noticed that this lovely crate does not have an implementation of the ITP method root solver, which is, quoting [Wikipedia](https://en.wikipedia.org/wiki/ITP_method):

"The ITP method, short for Interpolate Truncate and Project, is the first root-finding algorithm that achieves the superlinear convergence of the secant method while retaining the optimal worst-case performance of the bisection method. It is also the first method with guaranteed average performance strictly better than the bisection method under any continuous distribution. In practice it performs better than traditional interpolation and hybrid based strategies (Brent's Method, Ridders, Illinois), since it not only converges super-linearly over well behaved functions but also guarantees fast performance under ill-behaved functions where interpolations fail."

Practically, it is faster than `BrentRoot` due to fewer function evaluations. I thought this might be a handy contribution towards `argmin`, since I didn't see too many implementations of this online (There's the `itp` package in R, and in Julia I think it exists in `Roots.jl`. I also found it located within the Rust crate `kurbo` [here](https://github.com/linebender/kurbo/blob/main/src/common.rs#L646), but this was tuned specifically for curve fitting instead of generic root solving. I templated this implementation based off of `brentroot.rs`.

# Verification
I used the quadratic test from `brentroot.rs`, and it achieved a solution in one iteration faster. However, there's an additional function evaluation that occurs to satisfy the `argmin` iteration state, so currently comes in equal in terms of function evaluations with BrentRoot for that example specifically. 

Additionally, I added a test against the polynomial shown in the example on Wikipedia, and stepping through in a debugger I was able to verify that `a` and `b` bounds on the bracket (of the solver) indeed match what is in that table there.

Let me know what else should be changed, as I'm new to this community. Thanks! :smiley:

### DIFF
--- a/crates/argmin/src/solver/itp/itp_method.rs
+++ b/crates/argmin/src/solver/itp/itp_method.rs
@@ -27,6 +27,9 @@ pub enum ItpRootError {
     /// max must be larger than min
     #[error("ItpRoot error: max must be larger than min.")]
     MinLargerThanMax,
+    /// max and min are equivalent
+    #[error("ItpRoot error: max is equivalent to min.")]
+    MaxEqualToMin,
 }
 
 /// # ITP method
@@ -114,6 +117,10 @@ impl<F: ArgminFloat> ItpRoot<F> {
     /// kappa2 is defaulted to 2.0.
     /// n0 is defaulted to 1.0.
     pub fn from_defaults(min: F, max: F, tol: F) -> Result<Self, Error> {
+        if (max - min).is_zero() {
+            return Err(ItpRootError::MaxEqualToMin.into())
+        }
+
         Self::new(
             min,
             max,
--- a/crates/argmin/src/solver/itp/itp_method.rs
+++ b/crates/argmin/src/solver/itp/itp_method.rs
@@ -77,8 +77,20 @@ impl<F: ArgminFloat> ItpRoot<F> {
     /// The values `kappa1` and `kappa2` are hyper-parameters tuning the truncation size.
     /// The parameter `n0` is a hyper-parameter slack variable controlling the projection interval
     /// size.
-    pub fn new(min: F, max: F, tol: F, kappa1: F, kappa2: F, n0: F) -> Self {
-        ItpRoot {
+    pub fn new(min: F, max: F, tol: F, kappa1: F, kappa2: F, n0: F) -> Result<Self, Error> {
+        if tol < F::zero() {
+            return Err(ItpRootError::NegativeTol.into());
+        }
+        // This helps ensure the log evaluation (in the init) is stable
+        if min > max {
+            return Err(ItpRootError::MinLargerThanMax.into());
+        }
+        // It's important to check this to verify n1o2 doesn't panic
+        if tol.is_zero() {
+            return Err(ItpRootError::ZeroTol.into());
+        }
+
+        Ok(ItpRoot {
             tol,
             kappa1,
             kappa2,
@@ -87,10 +99,12 @@ impl<F: ArgminFloat> ItpRoot<F> {
             b: max,
             fa: F::nan(),
             fb: F::nan(),
+            // Starts at zero, increments in the solver
             j: F::zero(),
+            // These get computed in the solver
             n1o2: F::nan(),
             nmax: F::nan(),
-        }
+        })
     }
 
     /// Constructor with default hyperparameters
@@ -99,7 +113,7 @@ impl<F: ArgminFloat> ItpRoot<F> {
     /// kappa1 is defaulted to 0.2 / (max - min).
     /// kappa2 is defaulted to 2.0.
     /// n0 is defaulted to 1.0.
-    pub fn from_defaults(min: F, max: F, tol: F) -> Self {
+    pub fn from_defaults(min: F, max: F, tol: F) -> Result<Self, Error> {
         Self::new(
             min,
             max,
@@ -133,17 +147,6 @@ where
         self.fb = problem.cost(&self.b)?;
         if self.fa * self.fb > float!(0.0) {
             return Err(ItpRootError::WrongSign.into());
-        }
-        if self.tol < F::zero() {
-            return Err(ItpRootError::NegativeTol.into());
-        }
-        // This helps ensure the log evaluation is stable
-        if self.a > self.b {
-            return Err(ItpRootError::MinLargerThanMax.into());
-        }
-        // It's important to check this to verify n1o2 doesn't panic
-        if self.tol.is_zero() {
-            return Err(ItpRootError::ZeroTol.into());
         }
 
         // Preprocessing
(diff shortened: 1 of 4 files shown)


crates/argmin/src/solver/itp/itp_method.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// kappa1 is defaulted to 0.2 / (max - min).
+    /// kappa2 is defaulted to 2.0.
+    /// n0 is defaulted to 1.0.
+    pub fn from_defaults(min: F, max: F, tol: F) -> Self {
+        Self::new(
+            min,
+            max,
+            tol,
+            // kappa1, suggested from paper
+            float!(0.2) / (max - min),
```
duncanam, line 108: Wondering, maybe we want to have this function return `Result` and verify `max - min` cannot be zero so we don't panic here. Not sure though, maybe not worth it right now.

crates/argmin/src/solver/itp/itp_method.rs
```diff
@@ -0,0 +1,354 @@
...
+            self.b = xitp;
+        }
+        self.j = self.j + float!(1.0);
+
+        // Solver loop termination
+        if (self.b - self.a) <= (float!(2.0) * self.tol) {
+            let sol = (self.a + self.b) * float!(0.5);
+            // TODO: This function evaluation serves no purpose other than to serve argmin's cost
+            // method on the state. It feels wasteful.
+            let f_sol = problem.cost(&sol)?;
```
duncanam, line 209: This bothers me, but maybe someone smarter than me might know a clever way to not do this.

duncanam:
@stefan-k Does this seem like a worthwhile addition? If so, I did take a peek at CI here and some of these seem like flakes- I could totally be misreading, though.

stefan-k:
Hi @duncanam , this is definitely a worthwhile addition, thank you! Unfortunately it will take a bit until I can give a useful review, hopefully around the upcoming holidays. Apologies for being so unresponsive recently :/

duncanam:
> Hi @duncanam , this is definitely a worthwhile addition, thank you! Unfortunately it will take a bit until I can give a useful review, hopefully around the upcoming holidays. Apologies for being so unresponsive recently :/

No worries! Have a good holiday season!

crates/argmin/src/solver/itp/itp_method.rs
stefan-k: This sounds like a good addition!

crates/argmin/src/solver/itp/itp_method.rs
```diff
@@ -0,0 +1,354 @@
...
+            return Err(ItpRootError::NegativeTol.into());
+        }
+        // This helps ensure the log evaluation is stable
+        if self.a > self.b {
+            return Err(ItpRootError::MinLargerThanMax.into());
+        }
+        // It's important to check this to verify n1o2 doesn't panic
+        if self.tol.is_zero() {
+            return Err(ItpRootError::ZeroTol.into());
+        }
```
stefan-k, line 147: I know that `BrentRoot` also checks these fields in the `init` method of `Solver`, but I think it may be better to do these checks in the `new` method of `ItpRoot`. What do you think?

crates/argmin/src/solver/itp/itp_method.rs
```diff
@@ -0,0 +1,354 @@
...
+    /// The parameter `n0` is a hyper-parameter slack variable controlling the projection interval
+    /// size.
+    pub fn new(min: F, max: F, tol: F, kappa1: F, kappa2: F, n0: F) -> Self {
+        ItpRoot {
+            tol,
+            kappa1,
+            kappa2,
+            n0,
+            a: min,
+            b: max,
```
stefan-k, line 87: Question out of curiosity: Are there any sane defaults for these parameters (ideally something mentioned in the paper)?

crates/argmin/src/solver/itp/itp_method.rs
stefan-k: I agree. Do I understand correctly that the algorithm itself does not require to compute the final cost function value? It is only required here because otherwise the cost function value would not be related to the final solution?

stefan-k (commented):
Thank you for this PR! This is a very valuable addition and looks pretty good to me. :) I only have a few minor points to discuss. Regarding the math I'll have to trust you here as I currently am not able to dive into this in detail.

stefan-k:
I'll try to fix the CI problems in another PR.

crates/argmin/src/solver/itp/itp_method.rs
duncanam: Sure, sounds good. Should we move the logic in BrentRoot too to match code style, or keep the change isolated to this PR for now?

crates/argmin/src/solver/itp/itp_method.rs
duncanam: That was my goal with the defaults method in this file- is there a better way to notate that idiomatically? I've seen some packages use `default()`, but the "classmethod" style of using `from_` seems to also be clear.

duncanam:
> I agree. Do I understand correctly that the algorithm itself does not require to compute the final cost function value? It is only required here because otherwise the cost function value would not be related to the final solution?

(For whatever reason GitHub was disallowing a reply inline)

Yes- the solution architecture sets it up such that the subsequent evaluation would satisfy the tolerance, rather than needing to evaluate and check success criteria. We could, in theory, return maybe a theoretical value given this information, but it seems misleading.

crates/argmin/src/solver/itp/itp_method.rs
duncanam: Yes- the solution architecture sets it up such that the subsequent evaluation would satisfy the tolerance, rather than needing to evaluate and check success criteria. We could, in theory, return maybe a theoretical value given this information, but it seems misleading.

crates/argmin/src/solver/itp/itp_method.rs
duncanam: I moved the logic here at the very least. We can decide next on BrentRoot, although that would technically introduce a breaking change to downstream users as `new` would go from returning `Self` to `Result`

stefan-k: I'd keep it in `ItpRoot` only for now.

crates/argmin/src/solver/itp/itp_method.rs
```diff
@@ -0,0 +1,354 @@
...
+        if (self.b - self.a) <= (float!(2.0) * self.tol) {
+            let sol = (self.a + self.b) * float!(0.5);
+            // TODO: This function evaluation serves no purpose other than to serve argmin's cost
+            // method on the state. It feels wasteful.
+            let f_sol = problem.cost(&sol)?;
+            return Ok((
+                state
+                    .terminate_with(TerminationReason::SolverConverged)
+                    .param(sol)
+                    .cost(f_sol.abs()),
```
stefan-k, line 224: Is the `.abs()` necessary here?

crates/argmin/src/solver/itp/itp_method.rs
```diff
@@ -0,0 +1,354 @@
...
+                None,
+            ));
+        }
+
+        Ok((state.param(self.b).cost(self.fb.abs()), None))
```
stefan-k, line 229: Same question here regarding `.abs()`

duncanam:
I'm sorry this has taken so long. Let me try and pull this across the finish line.

PR #570, opened 2025-02-15
Fix licensing in book

GermanHeim:
Hello,
There were missing license links in the book. This should fix it.


stefan-k:
Thanks a lot @GermanHeim! Good catch. I fixed the CI in main, would you mind rebasing on main? Thanks!

GermanHeim:
Sorry, I merged by accident instead of rebasing! Is that fine?

stefan-k:
> Sorry, I merged by accident instead of rebasing! Is that fine?

I think so, I'll just squash the commits. Thanks for the quick response! :)

stefan-k (approved):

PR #575, opened 2025-03-08
Feature/update faer 0.21

geo-ant:
This adds support for `faer-rs` version 0.21.x, while keeping the previously added support for version 0.20. I realize, this looks like a daunting PR again in terms of lines changed, but it's actually not as bad. Since `faer` underwent a big rewrite between the versions, I had to factor out the actual implementations of the `argmin-math` traits into different modules and conditionally compile them. However, I also factored out the tests from the old `faer_m` module, which is now renamed to `faer_m_0_20`, to a common `faer_test` module. The tests were just factored out into this module, no changes were made to the tests*. That means the implementations for both faer backends pass the same, unaltered test suite.

In summary:
* I renamed `faer_m` (and it's submodules to) `faer_m_0_20`
* I factored out the tests in the submodules to `faer_test`
* I added `faer_m_0_21` for implementations in the 0.21 dependency

I hope newer `faer` versions will introduce less breaking changes for the fundamental operations we're using here, so that it won't be necessary to introduce `faer_m_x_xx` again. But even if not, this is reasonably simple to do.

(*) except at one place where I had to introduce a `matrix_element_at` helper function to access a matrix element across faer versions. But this is not a semantic change.


geo-ant:
urgh, the failing CI is related to https://github.com/argmin-rs/argmin/issues/574 , which is about the `paste` crate being unmaintained as of recently.

stefan-k:
Thanks for this PR, it's great to have the most recent version supported as well! :)

> urgh, the failing CI is related to #574 , which is about the `paste` crate being unmaintained as of recently.

"urgh" is the perfect noise for this. About 90% of the time I have available for argmin is spent on things like this ;) It should be fixed in main after merging #576

stefan-k (approved):

geo-ant:
> Thanks for this PR, it's great to have the most recent version supported as well! :)
> 
> 
> 
> > urgh, the failing CI is related to #574 , which is about the `paste` crate being unmaintained as of recently.
> 
> 
> 
> "urgh" is the perfect noise for this. About 90% of the time I have available for argmin is spent on things like this ;) It should be fixed in main after merging #576 

I feel you, and I hope you didn't take it as an insult to you. You have your CI set up really professionally, I'm going to take a page of your playbook for my projects at some point.

stefan-k:
> I feel you, and I hope you didn't take it as an insult to you. 

No no, not at all. I just felt like venting a little :D 

> You have your CI set up really professionally, I'm going to take a page of your playbook for my projects at some point.

Thanks and feel free to do so. Don't be surprised if you find a mess in certain corners ;)

PR #581, opened 2025-03-21
Render math in argmin-testfunctions with KaTeX (part 2)

airwoodix:
Follow-up to #530.

Render the math in `argmin-testfunctions` with KaTeX for the documentation that had not been converted yet. No code changes.

The function in `mccorminck.rs` seems to be call "McCormick" (without the `n`) on the Internet. See e.g. [wikipedia](https://en.wikipedia.org/wiki/Test_functions_for_optimization#:~:text=McCormick) and [here](https://www.sfu.ca/~ssurjano/mccorm.html). I used this spelling in the documentation, but didn't change the function or module name, to not introduce a breaking change in this documentation-only PR. @stefan-k Would you consider a PR changing the function and module names?

### DIFF
--- a/crates/argmin-testfunctions/src/eggholder.rs
+++ b/crates/argmin-testfunctions/src/eggholder.rs
@@ -5,29 +5,33 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-//! # Eggholder test function
+//! # Eggholder test function.
 //!
 //! Defined as
 //!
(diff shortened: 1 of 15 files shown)


stefan-k (approved):
Thanks so much, this is again absolutely excellent! :)

Oops, at least I was consistent with the typo! I'd appreciate it if you could fix this in a PR, even if it is a breaking change :)

stefan-k:
Pipeline should be fixed with #582

airwoodix:
Thanks @stefan-k for the feedback and the pipeline fix. I rebased this PR's branch onto `main`.

stefan-k:
@airwoodix Thanks again for the excellent work and apologies (again) for the long wait!

PR #583, opened 2025-03-29
Support `wasm32-unknown-unknown` without `wasm-bindgen`

rikhuijzer:
Thank you for making this software available. I think numerical packages in pure Rust are really important. 

This PR allows for succesfully building armin with
```sh
$ cargo build -p argmin --no-default-features --target wasm32-unknown-unknown
```
As a bit of background, the wasm (synonym: WebAssembly) situation is a bit complication right now. As I understand it, most wasm targets assume that Javascript is available too. So that's why `getrandom/js` often works. However, more and more wasm targets are moving away from Javascript. Wasmtime, for example, is a wasm runtime that does not support Javascript. This is useful for example for plugin systems like is used by Typst or Zellij. Both allow developers to write plugins which then can be used by users. But to not compromise the whole system, the plugins are sandboxed by default. So these plugins typically are not allowed to access the filesystem, or other system interfaces; also not via JS bindgen.

But I think it would be very nice if we could argmin in these barebone environments. To do so, this PR suggests to fix the build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,7 +203,9 @@ jobs:
       - name: Install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh 
       - name: Build target wasm32-unknown-unknown
-        run: cargo build --workspace --exclude argmin-observer-spectator --exclude spectator --exclude argmin-observer-paramwriter --exclude "example-*" --exclude argmin-testfunctions-py --target wasm32-unknown-unknown
+        run: cargo build --workspace --exclude argmin-observer-spectator --exclude spectator --exclude argmin-observer-paramwriter --exclude "example-*" --exclude argmin-testfunctions-py --target wasm32-unknown-unknown --no-default-features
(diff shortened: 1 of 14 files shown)


rikhuijzer:
@stefan-k My apologies for the failing CI jobs in [68cf721](https://github.com/argmin-rs/argmin/pull/583/commits/68cf721cc5f6c900bb646f08aeea2f4788424b71) and [dc43934](https://github.com/argmin-rs/argmin/pull/583/commits/dc4393485237363886b883518764f3498fdc4fec).

I now tested this PR in my own branch and CI should pass now. I also added a test that explicitly tests this `wasm32-unknown-unknown` without `wasm-bindgen`.

My compliments by the way for the CI testing setup. The tests are very thorough.

stefan-k:
@rikhuijzer Apologies that I didn't address this earlier. Strangely, the CI runs are gone and even more strangely, triggering the CI doesn't work either.

stefan-k:
@rikhuijzer Hooray, the CI still passes (more or less, I'll fix the clippy lints later). Thanks a lot for addressing this issue and apologies again for the long wait.

PR #608, opened 2025-09-12
Bump remaining dependencies

JosiahKane:
Dependabot seems to be struggling with the rebases, so this should just clear the rest of the set (https://github.com/argmin-rs/argmin/pull/555 https://github.com/argmin-rs/argmin/pull/589 https://github.com/argmin-rs/argmin/pull/535). 

The ndarray one (https://github.com/argmin-rs/argmin/pull/513) is a false alarm since (outside of explicit compatibility code) you're already on v16.


stefan-k (approved):
Thanks again! :)

PR #614, opened 2025-09-15
Update nalgebra to 0.34

JosiahKane:
These dependabot PRs just keep coming. I'm inclined to leave bincount and criterion: as development tools it shouldn't matter to consumers of argmin whether they're fully up to date. 
Still, if we're to get a release out we might as well get nalgebra. Someone is sure to come asking for it promptly if we miss it.
This is basically just copy-paste from the handling of 0.33. The tests seem happy and I didn't see anything to alarm me in the changelog. 

I think one interesting question for these updates is how you handle your minimum rust version. Nalgebra have increased their minimum versions with this to 1.87 (i.e the release from May this year). According to `cargo msrv`, the main argmin crate currently works at 1.78 which was May last year. If you have any sort of standard where you try to support the last X rust versions, this could break that.

Of course it shouldn't be a huge problem anyway: with the optional dependency machinery there should still be space for someone to use an older nalgebra version if they can't update rust.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -18,13 +18,13 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2025-02-25
+          toolchain: nightly-2025-09-14
 
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
 
       - name: Generate code coverage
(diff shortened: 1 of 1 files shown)


stefan-k:
I suspect dependabot has given up because none of its PRs were merged. Your recent efforts have woken it from its slumber!

Supporting the newest nalgebra is great, especially if it's possible with little changes.

I agree though that dev dependencies do not necessarily need to be updated.

Regarding the MSRV: I probably should, but so far don't really care. With the optional dependencies it's quite difficult to communicate under which conditions, which MSRV is valid.

Thanks again for the excellent work! :)

stefan-k (approved):

PR #617, opened 2025-09-20
Fix clippy complaint about manual implementations of is_multiple_of

JosiahKane:
This is pretty obvious. 
The newest Clippy believes, correctly, that it is easier to read `x.is_multiple_of(y)` than `x % y ==0`.

The only downside I can see is that with this change the msrv in argmin proper (rather than merely via sometimes-optional dependencies) is now [actually 1.87.0](https://doc.rust-lang.org/std/primitive.u32.html#method.is_multiple_of). You've already indicated that's not something you particularly care about, so this is definitely the cleanest way to resolve the clippy failure.


stefan-k (approved):
This is indeed easier to read, thanks!

The MSRV the is quite recent, but I guess (read: hope) that's fine.

PR #627, opened 2025-09-30
update faer to 0.23

Ravenslofty:
This is fortunately much shorter than #575 thanks to the infrastructure introduced there; we can just use the 0.21 backend with the new release.

(too bad I missed the release of 0.11.0)


stefan-k:
Thanks a lot @Ravenslofty ! Since there are so few changes necessary, do you mind adding support for 0.22 as well?

What I see missing at first glance is adding checks for the new library versions to the CI pipeline.

stefan-k (approved):
Thanks a lot @Ravenslofty ! LGTM :)

PR #641, opened 2025-10-17
Change default time to None in IterState

### LINKED ISSUE #640: time is always 0ns
With `argmin 0.11.0`, I noticed that the time displayed at the end of the optimization is always `0ns`.

Example:

```bash
particleswarm> cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.27s
     Running `D:\workspace\argmin\target\debug\example-particleswarm.exe`
OptimizationResult:
    Solver:        Particle Swarm Optimization
    param (best):  Particle { position: [3.5844283403304917, -1.8481265269644036], velocity: [4.682418545689039, 4.416573268635766], cost: 0.0, best_position: [3.5844283403304917, -1.8481265269644036], best_cost: 0.0 }
    cost (best):   0
    iters (best):  245
    iters (total): 10000
    termination:   Maximum number of iterations reached
    time:          0ns
```

While with `argmin 0.10.0`:

```bash
particleswarm>cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.24s
     Running `D:\workspace\argmin\target\debug\example-particleswarm.exe`
OptimizationResult:
    Solver:        Particle Swarm Optimization
    param (best):  Particle { position: [3.0, 2.0], velocity: [-4.473958059868867, -4.722233620327177], cost: 0.0, best_position: [3.0, 2.0], best_cost: 0.0 }
    cost (best):   0
    iters (best):  241
    iters (total): 10000
    termination:   Maximum number of iterations reached
    time:          1.3887014s
```

GermanHeim:
## 📝 Description

Updated the `IterState` struct to initialize the time field as `None` instead of `Some(Duration::ZERO)`. This should fix the timer showing as 0 ns, which was introduced in commit https://github.com/argmin-rs/argmin/commit/6492020c5510c0aa4a99955a55207dbbfdee8a64.

## 🔨 Changes Made

- Updated the `IterState` struct to initialize the time field as `None` instead of `Some(Duration::ZERO)`.
- Modified tests to take into account this implementation

## 🔗 Related Issues

Closes https://github.com/argmin-rs/argmin/issues/640

### DIFF
--- a/crates/argmin/src/core/state/iterstate.rs
+++ b/crates/argmin/src/core/state/iterstate.rs
@@ -1028,7 +1028,7 @@ where
     /// # assert_eq!(state.last_best_iter, 0);
     /// # assert_eq!(state.max_iters, u64::MAX);
     /// # assert_eq!(state.counts.len(), 0);
-    /// # assert_eq!(state.time.unwrap(), Duration::ZERO);
+    /// # assert!(state.time.is_none());
     /// # assert_eq!(state.termination_status, TerminationStatus::NotTerminated);
     /// ```
     fn new() -> Self {
@@ -1057,7 +1057,7 @@ where
             max_iters: u64::MAX,
             counts: HashMap::new(),
             counting_enabled: false,
-            time: Some(Duration::ZERO),
+            time: None,
             termination_status: TerminationStatus::NotTerminated,
         }
     }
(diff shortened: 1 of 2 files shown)


stefan-k:
Lovely, thanks! 
Have you tested how this affects the observers?

GermanHeim:
From what I can see, the only observer that actually uses time is the [observer spectator](https://github.com/argmin-rs/argmin/blob/main/crates/argmin-observer-spectator/src/observer.rs) (in observe_iter and handle_connection).

Currently, it works because it uses `.unwrap_or(std::time::Duration::from_secs(0)),`, but this approach isn’t ideal.

One option would be to change the `time` field in `Samples` from `Duration` to `Option<Duration>` and adjust the code accordingly. Does that sound reasonable to you?

PR #653, opened 2026-01-23
Implement Operations for Col in faer

cloudsftp:
Hi, this PR adds implementations of some operations for the Col vectors of the faer crate.
The operation implementations are mostly copied and adjusted from the code added in https://github.com/argmin-rs/argmin/pull/575.

The implemented operations are enough for the GaussNewton solvers.

### Why?

When implementing the Operator and Jacobian traits for some objective function, it is awkward to implement it with `type Param = Mat`.
The `Col` type is much better.
The solver needs some operations to be implemented for `Col`.

I wanted to ask first, whether these changes are wanted, before continuing to implement other operations for the `Col` type that are required by other solvers.


cloudsftp:
Depends on https://github.com/argmin-rs/argmin/pull/654

